PR #75, opened 2020-03-10
adjusts size of viewport for loading/unloading frames based on whethe…

benptc:
…r we are in powersave mode.

Unloading all frames that are slightly outside viewport is a bit too strict if you aren't trying to save performance (if you don't have a lot of frames loaded at once) - and if you move the camera too fast you sometimes see the frame flicker back into existence - so it overall feels smoother if you are more generous with the viewport bounds.

I thought about making the `additionalBuffer` inversely proportional to the number of frames in the scene, but I think this is a good starting point and also gives the user control over this behavior.

### DIFF
--- a/src/gui/ar/draw.js
+++ b/src/gui/ar/draw.js
@@ -1428,40 +1428,7 @@ realityEditor.gui.ar.draw.drawTransformed = function (visibleObjects, objectKey,
                 }
 
                 if (this.isLowFrequencyUpdateFrame) {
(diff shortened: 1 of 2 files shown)


benptc:
Ah, that's a good judgment call. I'll make the change.

benptc:
> At its current level of complexity this would be nice to have as a separate function, potentially an all-in-one realityEditor.gui.ar.positioning.canUnload

Followed your recommendation and created a canUnload function - let me know what you think.

PR #76, opened 2020-03-10
re-correct target downloader so it caches jpg as downloaded but still…

benptc:
… prioritizes adding target via DAT if the file is available.

Previous solution always triggered the addNewMarkerJPG if it had been previously downloaded, rather than choosing between addNewMarker (DAT) and addNewMarkerJPG.


benptc:
Hopefully this doesn't have any sneaky side effects down the road..

PR #77, opened 2020-03-11
More thoroughly keep track of caching in target downloader 

benptc:
Previous fixes were stop-gap, didn't actually keep track of which files had been previously downloaded/cached and which ones hadn't was leading to edge case bugs. This tries to fix it properly.


benptc:
I added a fairly involved system for attempting to redownload target data if a download fails, instead of just accepting that they don't exist. It will attempt each one 10 times with a 10 second delay between attempts. I've tested this by deleting the target.jpg file from the realityobjects before starting the server, letting the app run for awhile, and then adding that file back to the server - within 10 seconds the found object will change from red to black and you'll be able to recognize it.

PR #78, opened 2020-03-13
fix upside-down app to work even if orientation event triggers before…

benptc:
… setProjectionMatrix

If you previously started the app while holding the phone upside down, the projection matrix would get initialized incorrectly.

(Reported https://forum.spatialtoolbox.vuforia.com/t/tool-does-not-attach-when-camera-is-on-the-left-side/25/7)


valentinptc:
Looks good.

PR #81, opened 2020-03-31
adds temporary fix for instant image targets

benptc:
This fixes half of the problem.

Intercepts matrices detected by Vuforia before they get passed to visibleObjects, and multiplies those that use the instant image target API by a scale factor. Ideally instead of scaling the size by 3x we would adjust the position so that it's just 3x closer to the camera, but this is the wrong place in the pipeline to do it and I'm not sure where the right place is.

Separately adjusts the visibility distance for instant image targets to be 3x as far because they are treated as being 3x as far away. 

Thoughts?

### DIFF
--- a/src/app/callbacks.js
+++ b/src/app/callbacks.js
@@ -211,6 +211,31 @@ createNameSpace('realityEditor.app.callbacks');
 
             });
 
+            realityEditor.forEachObject(function(object, objectKey) {
+                if (typeof visibleObjects[objectKey] !== 'undefined') {
+                    // if it's a JPG instant target, correct the size so it matches as if it were DAT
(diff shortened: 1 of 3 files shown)


benptc:
This is the code I used to produce these examples for the tutorial: 

09-attach-pocket-tool-to-object

10-move-object-with-pocket-tools

benptc:
Unfixed, it thinks everything is really far away so the tools are *tiny* - cute, but impractical unless you are viewing from within a few inches away

unfixed-instant-image-targets-lowres

benptc:
OK I'll merge this in for now in case anyone starts playing around with it in the next couple days, but still exploring other solutions.

PR #83, opened 2020-04-03
Intro instructions

benptc:
implemented using the modal.js class I wrote for the git commit feature, but with different CSS styles and a header field.

Note: latest iOS app version is required for these links to open in safari instead of in-app

intro-screen-iphone-8plus

### DIFF
--- a/css/modal.css
+++ b/css/modal.css
@@ -162,8 +162,8 @@
     left: 0;
     width: 100vw;
     height: 100vh;
-    background-color: rgba(0,0,0,0.5);
-    -webkit-backdrop-filter: blur(10px);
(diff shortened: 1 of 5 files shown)


hobinjk-ptc:
Have you tried target=_blank? Not sure if there's other possible approaches

benptc:
> Have you tried target=_blank? Not sure if there's other possible approaches

I hadn't thought of that so I just tried that out. Without the change to the iOS project, target=_blank in the app for some reason just makes nothing happen when you click on the link. With the change to the iOS project, clicking on the link opens it in the external browser, as expected.

valentinptc:
All good all working.

PR #84, opened 2020-04-04
Allow server from different ports

valentinptc:
This is a critical update to allow servers with different ports.
It is required to improve stability on iOS devices.

Why?
Its possible that multiple server run in the background on an ios device. 
Using common ports like 80, 8080 or 8888 can increase the potential for application crashes.
Therefore the server is moved to a new port. That needs to be reflected in:
ios app
user interface
server

### DIFF
--- a/src/device/onLoad.js
+++ b/src/device/onLoad.js
@@ -302,7 +302,8 @@ realityEditor.device.onload = function () {
     // see if we should open the modal
     let shouldShowIntroModal = window.localStorage.getItem('neverAgainShowIntroTips') !== 'true';
 
-    if (shouldShowIntroModal) {
+    // if (shouldShowIntroModal) {
+    if (false) {
         let modalBody = "The Vuforia Spatial Toolbox is an open source research platform for exploring Augmented Reality and Spatial Computing.<br>" +
             "<ul>" +
             "</li>1. <a class='modalLink' href='https://spatialtoolbox.vuforia.com/docs/use/using-the-app'>Learn how to use the Spatial Toolbox</a><br><br></li>" +
--- a/src/device/onLoad.js
+++ b/src/device/onLoad.js
@@ -300,10 +300,10 @@ realityEditor.device.onload = function () {
     }
 
     // see if we should open the modal
-    let shouldShowIntroModal = window.localStorage.getItem('neverAgainShowIntroTips') !== 'true';
+    // let shouldShowIntroModal = window.localStorage.getItem('neverAgainShowIntroTips') !== 'true';
 
     // if (shouldShowIntroModal) {
-    if (false) {
+    if (false) { // eslint-disable-line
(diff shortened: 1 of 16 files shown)


src/device/onLoad.js
```diff
@@ -300,9 +300,10 @@ realityEditor.device.onload = function () {
     }
 
     // see if we should open the modal
-    let shouldShowIntroModal = window.localStorage.getItem('neverAgainShowIntroTips') !== 'true';
+    // let shouldShowIntroModal = window.localStorage.getItem('neverAgainShowIntroTips') !== 'true';
 
-    if (shouldShowIntroModal) {
+    // if (shouldShowIntroModal) {
+    if (false) { // eslint-disable-line
```
hobinjk-ptc, line 306: Make sure to revert this before merging

src/states.js
```diff
@@ -47,13 +47,14 @@
...
+/* exported httpPort, defaultPort, timeForContentLoaded, timeCorrection, boundListeners, TEMP_DISABLE_MEMORIES, globalStates, globalCanvas, publicDataCache, pocketItemId, globalScaleAdjustment, pocketFrame, pocketNode, globalDOMCache, shadowObjects, globalProgram, rotateX, rotationXMatrix, editingAnimationsMatrix, pocketDropAnimation, pocketBegin, visibleObjectTapInterval, visibleObjectTapDelay, overlayDiv, CRAFTING_GRID_WIDTH, CRAFTING_GRID_HEIGHT, DEBUG_DATACRAFTING */
 
 /**********************************************************************************************************************
  ******************************************** constant settings *******************************************************
  **********************************************************************************************************************/
 
 var httpPort = 8080;
+var defaultPort = '8080';
```
hobinjk-ptc, line 57: Is there a benefit to making this a string? It appears that other places specify ports as numbers instead

hobinjk-ptc: Notably this has the potential to break things like `myPort === otherPort` if those ever crop up

src/device/onLoad.js
valentinptc: I put this in on purpose, so that the Help screen does not show up. Apparently it was not a problem after all. I want to find a nicer method to show a little question mark icon once your not touching anything on the screen after you start the app. For as long as we don't have this method, the start screen is removed and therefore this if statement is just false.

src/states.js
valentinptc: > Is there a benefit to making this a string? It appears that other places specify ports as numbers instead

It does not make a difference, as its always used as part of a string and therefore the number is translated to string. Since it is always used as part of a string, I saved it as string.

valentinptc: > Notably this has the potential to break things like `myPort === otherPort` if those ever crop up

fixed

src/device/onLoad.js
hobinjk-ptc: Oh, then it would be much nicer to just do `let shouldShowIntroModal = false; // old code`

PR #85, opened 2020-04-08
adds port as argument to startVideoRecording js<>objc api

benptc:


src/states.js
```diff
@@ -54,7 +54,7 @@
  **********************************************************************************************************************/
 
 var httpPort = 8080;
-var defaultHttpPort = '8080';
+var defaultHttpPort = 8080;
```
benptc, line 57: I think this is safe to do? Seems better to keep all ports as numbers

hobinjk-ptc: 👍

PR #97, opened 2020-05-22
Fixes groundplane snapping to world origin

benptc:
Previous snapping relied on a global worldCorrection variable, but now each world has an origin that can be accessed, so uses that instead.

Also adds a definition for sliders in the settings menu, and ignores right clicks, but so that those features can be used in content_script addons.


css/index.css
```diff
@@ -877,6 +877,19 @@ body {
...
+#settingsIframe {
+    background-color: #222222;
+    position: absolute;
+    left: 0;
+    top: 0;
+    bottom: 0;
+    right: 0;
+    height: 100%;
+    width: 100%;
+    -webkit-transform: translateZ(9000px);
```
hobinjk-ptc, line 889: Would be nice to also have the non-prefixed version of this for redundancy

PR #99, opened 2020-05-28
"environment" refactor

benptc:
Can I get feedback on this design change?

**Motivation**: the iPad rendering was broken because it mistakenly evaluated isDesktop to true because of changes to window.navigation.userAgent. Rather than simply change this to a more accurate check, I thought it would be better if some variables/features were explicitly enabled by the add-ons that require them (e.g. internalResearch-addon), and default to normal app behavior if the add-on is not present.

**Method**: I split up the isDesktop variable into a set of variables based on what it was being used for, and added setters/getters for those capabilities to src/device/environment.js. This means that new add-ons can selectively trigger certain behavior in the core app by toggling these environment vars. Essentially decouples the add-on from the implementations of the capabilities it requires.

Corresponding changes made to addon, e.g. https://github.com/ptcrealitylab/vuforia-spatial-internalResearch-addon/pull/2

### DIFF
--- a/src/device/environment.js
+++ b/src/device/environment.js
@@ -1,0 +1,14 @@
+createNameSpace("realityEditor.device.environment");
+
+(function(exports) {
+
+    /**
+     * Public init method sets up module and registers callbacks in other modules
+     */
+    function initService() {
+
+    }
+
+    exports.initService = initService;
+
+}(realityEditor.device.environment));
--- a/src/device/environment.js
+++ b/src/device/environment.js
@@ -33,6 +33,14 @@ createNameSpace("realityEditor.device.environment");
 
     exports.supportsDistanceFading = function() {
         return true; // false for desktop?
-    }
+    };
+
+    exports.getDistanceScaleFactor = function() {
+        return 1; // 10 for desktop?
+    };
+
+    exports.shouldCreateDesktopSocket = function() {
+        return false; // true for desktop
+    };
 
 }(realityEditor.device.environment));
--- a/src/device/environment.js
+++ b/src/device/environment.js
@@ -1,46 +1,110 @@
 createNameSpace("realityEditor.device.environment");
 
+/**
+ * @fileOverview realityEditor.device.environment.js
+ * This provides an extensible location for defining environment variables that are used by
+ * the core application, but may be modified by add-ons to affect conditional behavior in the app.
+ *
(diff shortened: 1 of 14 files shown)


src/device/environment.js
```diff
@@ -0,0 +1,110 @@
...
+        requiresMouseEvents: false,
+        supportsDistanceFading: true,
+        shouldCreateDesktopSocket: false,
+        // numbers
+        lineWidthMultiplier: 1, // 5
+        distanceScaleFactor: 1 // 10
+    };
+
+    // variables can be directly set by add-ons by using the public 'variables' property
+    exports.variables = variables;
```
hobinjk-ptc, line 37: If you want this to be pre-deprecated you could give it a "hidden" name like `exports._variables` or `exports.__variables`

src/device/utilities.js
```diff
@@ -290,16 +290,13 @@ realityEditor.device.utilities.diffArrays = function(oldArray, newArray) {
...
-realityEditor.device.utilities.isItADesktop = window.navigator.userAgent.indexOf('Mobile') <= -1 || window.navigator.userAgent.indexOf('Macintosh') > -1;
-
 /**
  * Utility function to determine if the editor is loaded on a desktop browser or within the full mobile app.
  * @todo: make more robust so that loading on a mobile safari browser is distinguisable from within the Reality Editor app
  * @return {boolean}
  */
 realityEditor.device.utilities.isDesktop = function() {
-    return realityEditor.device.utilities.isItADesktop;
+    return false;
```
hobinjk-ptc, line 299: This can just be removed with the new env system, right?

benptc: good catch - will remove!

src/device/environment.js
benptc: I'm not sure if I understand pre-deprecation. Right now I'm using that variable publicly to set its values in the addon, for instance
```
let env = realityEditor.device.environment;
env.variables.providesOwnUpdateLoop = true;
env.variables.requiresMouseEvents = true;
```

Is it bad practice to set it this way, but encourage/require the getter functions for getting the value?

hobinjk-ptc: If you're using it that way then I agree with the current name. From the comment that I can't find anymore it sounded like you were planning to only ever have setters/getters and never use .variables

PR #101, opened 2020-06-01
Anchor Objects feature + "where is?" navigation merge

benptc:
Attempting to merge back into master the anchor feature and the "Where Is?" feature that were developed in tandem.

Notable changes to look out for:
- All world objects require localization (you can't see their content or attach tools to them until you look at the image/etc target.
- The target downloader process is able to determine if an anchor should be created by looking at its checksum and its isAnchor property
- Some new math for repositioning anchors in space, but the rendering is handled by the regular update loop. Visible anchors get injected into the visibleObjects list prior to rendering.
- new src/gui/spatial module for rendering simple 2d navigation arrows to objects or tools. Can be triggered from server web interface, or a button in the Found Objects page

anchor-demo-with-tools 2020-06-01 16_59_42


src/network/index.js
```diff
@@ -861,6 +861,25 @@ realityEditor.network.onAction = function (action) {
     if (thisAction.loadLogicIcon) {
         this.loadLogicIcon(thisAction.loadLogicIcon);
     }
+
+    // Set states to locate object in space
+    if (thisAction.spatial) {
+        if(thisAction.spatial.locator){
```
hobinjk-ptc, line 867: nit: pad if statement parts

benptc: Thanks for the review!

PR #111, opened 2020-06-26
Add Image Target Upload

ddgond:
Adds the ability to upload an image target to an anchor object on the server directly from the app.

### DIFF
--- a/src/gui/settings/foundObjects.html
+++ b/src/gui/settings/foundObjects.html
@@ -108,6 +108,9 @@
         <div class="card reload-notice" hidden>
             <p class="content-padded">Please reload the app after you have finished uploading targets.</p>
         </div>
+        <div class="card error-notice" hidden>
+            <p class="content-padded">Please reload the app after you have finished uploading targets.</p>
+        </div>
         <div class="card">
             <ul class="table-view" id="objectList"></ul>
         </div>
@@ -192,6 +195,9 @@
             return;
         }
         
+        // Currently, there's a bug where adding an image target to an anchor replaces the id of the object, meaning
+        // that the UI can't tell that the new object it detected after upload is the same as the anchor. ignoredAnchors
+        // detects 'outdated' anchors and makes sure not to display them.
         var ignoredAnchors = [];
         var msgIds = Object.keys(msg);
         for (var i = 0; i < msgIds.length - 1; i++) {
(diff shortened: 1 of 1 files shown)


src/gui/settings/foundObjects.html
```diff
@@ -302,6 +353,40 @@
...
+          if (xhr.status === 200) {
+              // File(s) uploaded.
+              console.log('successful upload');
+              button.innerText = "Success!";
+              document.querySelector('.reload-notice').removeAttribute('hidden');
+              setTimeout(function() {
+                button.innerText = "Upload Target";
+              }, 3000);
+          } else {
+              console.log('error uploading');
```
hobinjk-ptc, line 379: It would be good for this to bubble up the same way as the success message, ideally with an explanation of the error

src/gui/settings/foundObjects.html
```diff
@@ -95,6 +105,9 @@
 
     <!-- Wrap all non-bar HTML in the .content div (this is actually what scrolls) -->
     <div class="content" style="right:65px;">
+        <div class="card reload-notice" hidden>
+            <p class="content-padded">Please reload the app after you have finished uploading targets.</p>
+        </div>
```
benptc, line 110: I like the visual feedback you added here!

src/gui/settings/foundObjects.html
```diff
@@ -165,16 +184,34 @@
      * This gets triggered by a POST message coming into this iframe from the parent window.
      * It receives a set of information for each discovered object, and renders a table displaying the information.
      * This should get called on a (infrequent) loop as long as the iframe is open, so that it doesn't keep stale data.
-     * @param {Object.<string, {name: string, version: string, ip: string, initialized: boolean, frames: Object, isAnchor: boolean, isWorld: boolean}>} msg
+     * @param {Object.<string, {name: string, version: string, ip: string, port: string, initialized: boolean, frames: Object, isAnchor: boolean, isWorld: boolean}>} msg
```
benptc, line 187: Thanks for updating

src/gui/settings/foundObjects.html
```diff
@@ -165,16 +184,34 @@
...
      */
     realityEditor.gui.settings.getObjectsFunction = function (msg) {
         // don't re-render if nothing has changed
         if (previousFoundObjects && JSON.stringify(previousFoundObjects) === JSON.stringify(msg)) {
             return;
         }
+        
+        var ignoredAnchors = [];
```
benptc, line 195: Can you add a short comment about why we have ignoredAnchors? Is this because the object's ID changes when we upload a new target image? We should eventually fix that on the server but this works in the short-term..

PR #121, opened 2020-07-10
iPhone notch fix

### LINKED ISSUE #110: Menu buttons need offset on iPhone XS
The "notched" screen of iPhone XS covers up part of the sidebar buttons. Make sure they get properly repositioned for each phone model with that screen shape.

ddgond:
Resolves #110 
iPhone XS previously wasn't included among phones with notches and iPhone 11 miscalculated the notch size.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -665,9 +665,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.15",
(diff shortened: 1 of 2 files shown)


benptc:
I just pushed a commit that rebases this based on the latest master branch. I'll merge it.

PR #133, opened 2020-08-07
Block categories

dangond-ptc:
Fixes bug that allows block names to be properly displayed in different categories


hobinjk-ptc:
Superseded by #136

PR #138, opened 2020-10-01
fixed an issue where p5 takes resources even if its not in use.

valentinptc:
This fix makes sure that the p5 loop, webgl context and very much all p5 is removed when not in use and reloads once required.

### DIFF
--- a/src/gui/spatial/index.js
+++ b/src/gui/spatial/index.js
@@ -335,7 +335,7 @@ realityEditor.gui.spatial.sketch = function(p) {
         /*  for(let key in this.nodeList) {
               this.draw.nodesP5(this.nodeList[key],p);
           }*/
-        
+
         for(let key in this.whereIsList) {
             this.draw.whereIsP5(this.whereIsList[key],p);
         }
--- a/src/gui/spatial/index.js
+++ b/src/gui/spatial/index.js
@@ -40,7 +40,7 @@ realityEditor.gui.spatial.lineAnimationList = {};
 realityEditor.gui.spatial.checkState = function() {
     realityEditor.gui.spatial.historianOn = false;
     realityEditor.gui.spatial.spatialOn = false;
-    
+
     for (let ip in globalStates.spatial.whereIs) {
         if (Object.keys(globalStates.spatial.whereIs[ip]).length > 0) {
             realityEditor.gui.spatial.spatialOn = true;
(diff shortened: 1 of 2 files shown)


src/gui/spatial/index.js
```diff
@@ -35,10 +37,80 @@ realityEditor.gui.spatial.clearSpatialList = function (){
 };
 realityEditor.gui.spatial.lineAnimationList = {};
 
+
+
+
+realityEditor.gui.spatial.checkState = function() {
+    let counter = 0;
+    let counter2 = 0;
+    for (let ip in globalStates.spatial.whereIs) {
```
hobinjk-ptc, line 46: Since you only care whether counter and counter2 are greater than 0 this code can be significantly optimized by replacing them with booleans historianOn and spatialOn then modifying all these for loops as follows:
```javascript
let spatialOn = false;
let historianOn = false;
for (let ip in globalStates.spatial.whereIs) {
  if (Object.keys(globalStates.spatial.whereIs[ip]).length > 0) {
    spatialOn = true;
    break;
  }
}

if (!spatialOn) {
  for (let ip in globalStates.spatial.howFarIs) {
    if (Object.keys(globalStates.spatial.howFarIs[ip]).length > 0) {
      spatialOn = true;
      break;
    }
  }
}
```

src/gui/spatial/index.js
```diff
@@ -291,7 +339,7 @@ realityEditor.gui.spatial.sketch = function(p) {
         /*  for(let key in this.nodeList) {
               this.draw.nodesP5(this.nodeList[key],p);
           }*/
-
+
```
hobinjk-ptc, line 342: nit: random whitespace

hobinjk-ptc (changes requested):

hobinjk-ptc:
Great to have these changes! One comment inline with a faster way of writing the spatialOn/historianOn checks

src/gui/spatial/index.js
```diff
@@ -35,10 +37,76 @@ realityEditor.gui.spatial.clearSpatialList = function (){
...
+    realityEditor.gui.spatial.spatialOn = false;
+    
+    for (let ip in globalStates.spatial.whereIs) {
+        if (Object.keys(globalStates.spatial.whereIs[ip]).length > 0) {
+            realityEditor.gui.spatial.spatialOn = true;
+            break;
+        }
+    }
+
+    for (let ip in globalStates.spatial.howFarIs) {
```
hobinjk-ptc, line 51: This is better, but if you want it to be really optimized this loop should be skipped if the previous loop set spatialOn to true, i.e. this:
```javascript
if (!realityEditor.gui.spatial.spatialOn) {
  for (let ip in globalStates.spatial.howFarIs) {
      if (Object.keys(globalStates.spatial.whereIs[ip]).length > 0) {
            realityEditor.gui.spatial.spatialOn = true;
            break;
      }
  }
}
```

PR #147, opened 2021-02-24
Fixes for screen extension with scene graph

benptc:
Updates the calculations for converting tool positions between screen and AR spaces. Shouldn't affect app behavior unless interacting with the screens-addon (yet to be published).

Also fixes a bug with the scene graph not processing/rendering tools when you lift them away from one object and carry them away from that object.

The initial position of tools when they pop out of the screen is occasionally buggy but I'll follow up with another commit to fix that before publishing the addon.

Also swapped out p5.js (4.6 MB) with p5.min.js (614 KB) to improve load time.


src/gui/screenExtension.js
```diff
@@ -44,6 +44,61 @@ realityEditor.gui.screenExtension.registerCallback = function(functionName, call
...
+    
+    // TODO: don't enable if this has a certain environment variable
+
+    // register screen objects when they are loaded
+    realityEditor.network.addObjectDiscoveredCallback(function(object, objectKey) {
+        // check if this object is visualization===screen and if so, add to registeredScreenObjects
+        if (object.visualization !== 'screen') { return; }
+
+        realityEditor.gui.screenExtension.registeredScreenObjects[objectKey] = {
+            object : objectKey,
```
hobinjk-ptc, line 62: nit: random spaces before :

hobinjk-ptc (approved):
Great work!

dangond-ptc (approved):
Looks good to me!

PR #148, opened 2021-02-25
checks that tool exists before processing messages coming from the…

benptc:
…iframe - fixes a bug if the uuid of the tool changes in between messages being sent


afustePTC (approved):
Perfect thanks Ben!

dangond-ptc (approved):

PR #155, opened 2021-03-12
Gl perf improvements

hobinjk-ptc:
Removes an unused memory hog of a debug variable and cleans up a bit of code


benptc (approved):
I'm admittedly unqualified to approve this but looks good to me?

PR #159, opened 2021-03-24
loads node properties correctly from server when initNode is called from new tool

benptc:
fixes bug where type-specific node properties (such as invisibility and default publicData fields) didn't propagate to app when initNode is called. if server sends back the initialized node data as a response to the post request, the app now uses that to update the local placeholder node that it generates.


afustePTC (approved):
Great

PR #165, opened 2021-05-26
Added aspect ratio control for demos

dangond-ptc:


hobinjk-ptc (changes requested):
I think this should also toggle off the notch handling code. The padding ends up looking strange since the adjusted webview doesn't get covered by the notch

hobinjk-ptc (approved):
Tested locally and LGTM. This is great to have! One minor paper cut here is that since it reloads the UI the toggle switch is in an inconsistent state and doesn't work to disable the feature. Since this is a developer feature I don't consider it a blocker but it would be nice to fix.

PR #169, opened 2021-06-22
"Human" objects and improved realtime for remote operator

benptc:
Adds a new `humanObjects.js` module that tells the server to create a humanObject for this device when the device localizes within that server's world object. Updates the `realtime.js` system to batch and send updates more efficiently. When the realtime mode is enabled, constantly updates the humanObject's position on the server's sceneGraph using these improved socket messages. The object's position will match the phone camera's position relative to the world.

Additionally, tweaks some settings for the remote operator.


dangond-ptc (approved):
Looks good to me!

PR #183, opened 2021-11-05
Use memory warning API to stop area target scan before crash

benptc:
When toggled on in the develop menu, stop the area target scan when app memory exceeds a specified percentage of total device memory (default = 33%). Also responds to native `UIApplicationDidReceiveMemoryWarningNotification` but that usually happens too late in my tested environment (around 50%) to upload without crashing. 

Native APIs already exist, no other PRs needed for this change.

### DIFF
--- a/src/gui/ar/areaTargetScanner.js
+++ b/src/gui/ar/areaTargetScanner.js
@@ -369,14 +369,6 @@ createNameSpace("realityEditor.gui.ar.areaTargetScanner");
         }
 
         feedbackString = status + '... (' + statusInfo + ')';
-
-        // if (globalStates.debugSpeechConsole) {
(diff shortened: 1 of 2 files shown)


src/gui/ar/areaTargetScanner.js
```diff
@@ -59,7 +62,7 @@ createNameSpace("realityEditor.gui.ar.areaTargetScanner");
             }
 
             // check if it's a world object
-            if (object && (object.isWorldObject || object.type === 'world')) {
+            if (object && !object.deactivated && (object.isWorldObject || object.type === 'world')) {
```
hobinjk-ptc, line 65: Nice! This is a great change in behavior to have

hobinjk-ptc (approved):

PR #189, opened 2021-11-17
Metaverse demo bugfixes

benptc:
- fix keyboard event propagation (in conjunction with PR for remote operator add-on)
- surface anchors don't conflict with settings menu scrolling
- correctly post tool positions to server when moved with surface anchors
- option to hide the origin cube


hobinjk-ptc (approved):
Great changes to have!

PR #191, opened 2021-11-29
Add fancy mesh animation

hobinjk-ptc:


src/gui/threejsScene.js
```diff
@@ -362,92 +376,220 @@ import { BufferGeometryUtils } from '../../thirdPartyCode/three/BufferGeometryUt
...
-          `
+        areaTargetVertexShader(center) {
+            return THREE.ShaderChunk.meshphysical_vert
+                .replace('#include <worldpos_vertex>', `#include <worldpos_vertex>
+    vPosition = position;
+    len = length(vPosition - vec3(${center.x}, ${center.y}, ${center.z}));
+    vWorldPosition = (modelMatrix * vec4( transformed, 1.0 )).xyz;`).replace('#include <common>', `#include <common>
+    varying vec3 vPosition;
+    varying float len;
+    varying vec3 vWorldPosition;`);
```
benptc, line 389: Oh, I was wondering how to edit an existing three.js shader instead of starting from scratch, this is pretty neat

src/gui/threejsScene.js
```diff
@@ -362,92 +376,220 @@ import { BufferGeometryUtils } from '../../thirdPartyCode/three/BufferGeometryUt
...
-                    "time": {value: 1.0},
-                    "sineTime": {value: 1.0},
-                    "map": { value: sourceTexture },
-                    "maxHeight": {value: maxHeight}
-                },
-                vertexShader: this.areaTargetVertexShader(),
-                fragmentShader: this.areaTargetFragmentShader(),
-                side: THREE.FrontSide
-            });
+        areaTargetMaterialWithTextureAndHeight(sourceMaterial, maxHeight, center, animateOnLoad, inverted) {
```
benptc, line 550: That's a good idea to accept a source material rather than just the texture

src/gui/threejsScene.js
```diff
@@ -362,92 +376,220 @@ import { BufferGeometryUtils } from '../../thirdPartyCode/three/BufferGeometryUt
...
+            }
+            let dt = now - this.lastUpdate;
+            this.lastUpdate = now;
+
             let indicesToRemove = [];
             this.materialsToAnimate.forEach(function(entry, index) {
                 let material = entry.material;
                 if (entry.currentHeight < entry.maxHeight) {
-                    entry.currentHeight += entry.animationSpeed;
+                    entry.currentHeight += entry.animationSpeed * dt;
```
benptc, line 592: always a nice touch

benptc (approved):
Very cool to see the implementation! Sure to wow the audience 😄

src/gui/threejsScene.js
hobinjk-ptc: Thanks! I should do this to the fragment shader as well so it's a bit more robust to changes in the THREE version but this is good enough for now

PR #197, opened 2022-01-03
Misc UI improvements & support for new remote operator features

benptc:
- provides API that remote operator can access to filter out which objects get prioritized for adding new tools to them
- provides API that remote operator can access to trigger the scaling feature without pinching with 2 fingers
- small bugfix for new pocket UI
- hides groundplane anchors for tools hidden inside closed envelopes
- adds pointer move events to pocket to detect if mouse hovers over a tool icon (for mouse-based devices .e.g desktop)

### DIFF
--- a/src/device/index.js
+++ b/src/device/index.js
@@ -1211,8 +1211,6 @@ realityEditor.device.onDocumentMultiTouchMove = function (event) {
                     x: event.pageX,
                     y: event.pageY
                 }
-                console.log('center: ', centerTouch);
-                console.log('outer: ', outerTouch);
                 realityEditor.gui.ar.positioning.scaleVehicle(activeVehicle, centerTouch, outerTouch);
 
             } else {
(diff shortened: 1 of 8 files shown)


src/device/index.js
```diff
@@ -1182,65 +1187,87 @@ realityEditor.device.onDocumentMultiTouchMove = function (event) {
...
+                        y: event.touches[0].pageY
+                    };
+                    outerTouch = {
+                        x: event.touches[1].pageX,
+                        y: event.touches[1].pageY
+                    };
+                } else {
+                    // if you have two fingers on the screen (one on the frame, one on the canvas)
+                    // make sure the scale event is centered around the frame
+                    [].slice.call(event.touches).forEach(function(touch){
```
hobinjk-ptc, line 1249: Array.from() is more standard than [].slice.call

hobinjk-ptc (approved):
Nice! Excited for the remote operator changes enabled by this

src/device/index.js
benptc: Thanks for the tip! I sprinkled a few of those in awhile ago – I'll go ahead and swap them all out with Array.from while I'm at it

PR #204, opened 2022-02-23
Better touch event handling

hobinjk-ptc:
Paired with the latest changes on the remote operator's depth-representation branch this tries to do away with the touchEventCatcher pattern

It is a bit flaky and prone to race conditions with how convoluted the startup process is. I'm trying to improve this but if it's not working just click into and out of logic mode and the settings menu

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -811,7 +811,7 @@ body {
     left: 0;
     top: 0;
     background: transparent;
-    pointer-events: none;
+    pointer-events: auto;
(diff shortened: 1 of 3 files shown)


benptc (approved):
Seems reasonable, hopefully it all just works. I know I already sometimes run into race conditions when I have too many threejs tools in the scene at once, due to how I programmed their touch propagation, so that's something to keep an eye on too.

PR #209, opened 2022-04-07
3D video recording button

benptc:
When in virtualizer mode, a small record button pops up that can trigger 3D video recording APIs.

In future, this button could also be removed and the APIs could be triggered by tools added to the space.

Requires https://github.com/ptcrealitylab/vuforia-spatial-toolbox-ios-swift/pull/54 and https://github.com/ptcrealitylab/vuforia-spatial-remote-operator-addon/pull/17

IMG_0024

IMG_0025


css/index.css
```diff
@@ -1675,3 +1675,14 @@ Logic Interface CSS
     opacity: 0.3;
     pointer-events: none;
 }
+
+#recordPointCloudsButton {
+    position: absolute;
+    width: 40px;
+    height: 40px;
+    left: 20px;
+    top: calc(100vh - 20px - 40px);
```
dangond-ptc, line 1684: I believe this can just be `bottom: 20px` when using position:absolute

dangond-ptc (approved):
Looks good to me!

hobinjk-ptc:
One minor nitpick that would be nice to have is asking the server whether it can support recording and only showing the button if it can (or showing an error if startRecording is sent to a server that doesn't support it). In general I think the architecture of this part can be simplified by communicating directly with the server instead of going through the app.

css/index.css
benptc: thanks, updated!

benptc:
> One minor nitpick that would be nice to have is asking the server whether it can support recording and only showing the button if it can (or showing an error if startRecording is sent to a server that doesn't support it). In general I think the architecture of this part can be simplified by communicating directly with the server instead of going through the app.

Updated to directly communicate to server. Still need to think about best way of asking the server for its capabilities, but I do agree with that feedback.

benptc:
Thanks for the detailed reviews, by the way!

hobinjk-ptc:
Changes LGTM! It does feel like rzvIo could be important enough to be refactored out but that sounds like a future target (likely after it gets swapped to something ToolSocket-based for the cloud proxy to be happier)

benptc:
Sounds good! I had a similar thought about rzvIo but I agree that we can do that in future. I can log the deferred work in Issues so it doesn't get lost.

PR #217, opened 2022-04-26
Simplify loading effect shaders; Rewrite fragment shader as find-and-replace

hobinjk-ptc:
This uses the approach from the vertex shader in the fragment shader where it looks for parts of the three.js shader that probably won't change and puts the new code next to them. Previous approach was to copy-paste the entire three.js shader and add stuff to it willy-nilly which isn't as scalable. Both approaches can randomly explode when updating three.js but the replacement has the advantage of being much simpler.


benptc (approved):
Nice simplification!

PR #218, opened 2022-05-06
ProjectedZ PointerEvents API

benptc:
If there is an occlusion mesh added to the scene (on AR devices), pointerevents posted into iframes will contain an additional `projectedZ` property calculated by raycasting against the occlusion mesh. If there isn't an occlusion mesh, but there is an area target gltf (on remote operator), raycasts against that instead.

**Question for reviewers**: are we ok accepting a performance cost on AR devices of loading the area target occlusion gltf into the three.js scene for this feature? I expect we'll make more and more use of it in the future, if we also implement snapping tools' positions to surfaces, etc.

### DIFF
--- a/src/app/targetDownloader.js
+++ b/src/app/targetDownloader.js
@@ -91,10 +91,11 @@ createNameSpace("realityEditor.app.targetDownloader");
         const objectID = evt.data.objectID;
         window.localStorage.setItem(`realityEditor.navmesh.${objectID}`, JSON.stringify(navmesh));
 
-        // Occlusion removed in favor of distance-based fading, but could be re-enabled in the future
-        // let object = realityEditor.getObject(objectID);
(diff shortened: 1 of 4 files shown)


dangond-ptc (approved):
Looks good to me, but I wonder if it would be better to pass the x,y,z of the intersection rather than the distance from the camera, because I think most use cases will want to use the location rather than the depth value.

dangond-ptc (approved):
Looks good to me!

hobinjk-ptc:
Do you think it's possible to add a way to move a small amount off from the intersection point? If we're directly on the surface of the mesh it'll look a bit bad in the remote operator view. I think adding the ray used for the raycast would be enough but I'm not sure how many layers of matrix nonsense are going on

benptc:
Oh that's also a good idea! There's just one matrix multiplication for now, where I convert from absolute scene coordinates to ground plane coordinates, so I feel like applying that to the ray as well should allow for this.... I'll test...

hobinjk-ptc (approved):
Awesome enhancement! Can't wait to see this in use

PR #221, opened 2022-05-26
adds environment variables and callbacks for customized area target scanning

benptc:
necessary for the pop-up-onboarding-addon to adjust the UI and trigger a custom area target scanning workflow

Includes:
- `suppressObjectDetections`: so that we have the option to ignore UDP heartbeats while app is in main menu
- `suppressObjectRendering`: so that we can ignore all rendering and hide all AR elements when we open various menus, or during the scanning process
- `overrideAreaTargetScanningUI`: so that we can customize the buttons/text that appears while scanning
- `automaticallyPromptForAreaTargetCapture`: can be disabled to ignore the default scanning flow
- callbacks for areaTargetScanner.js (`onStartScanning`, `onCaptureStatus`, `onStopScanning`, `onCaptureSuccessOrError`), and public method to `programmaticallyStartScan`: so that we can provide a customized UI/UX for scanning
- also adds a progress bar that updates while the area target is being generated!

### DIFF
--- a/src/gui/ar/areaTargetScanner.js
+++ b/src/gui/ar/areaTargetScanner.js
@@ -33,7 +33,7 @@ createNameSpace("realityEditor.gui.ar.areaTargetScanner");
         onCaptureStatus: [],
         onStopScanning: [],
         onCaptureSuccessOrError: []
-    }
+    };
 
     /**
      * Public init method to enable rendering ghosts of edited frames while in editing mode.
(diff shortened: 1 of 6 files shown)


src/gui/ar/areaTargetScanner.js
```diff
@@ -28,6 +28,13 @@ createNameSpace("realityEditor.gui.ar.areaTargetScanner");
     let limitScanRAM = false; // if true (toggled through menu), stop area target capture when device memory usage is high
     let maximumPercentRAM = 0.33; // the app will stop scanning when it reaches this threshold of total device memory
 
+    let callbacks = {
+        onStartScanning: [],
+        onCaptureStatus: [],
+        onStopScanning: [],
+        onCaptureSuccessOrError: []
+    }
```
hobinjk-ptc, line 36: nit: missing semicolon

src/gui/ar/areaTargetScanner.js
```diff
@@ -617,8 +708,34 @@ createNameSpace("realityEditor.gui.ar.areaTargetScanner");
         }
     }
 
+    exports.getDetectedServers = function() {
+        return detectedServers;
+    }
```
hobinjk-ptc, line 713: nit: missing semicolon

hobinjk-ptc (approved):

PR #222, opened 2022-05-26
adds rayDirection to worldIntersectPoint

benptc:
I realized that I didn't include this in the previous version, but you might want easy access to the `rayDirection` within the `worldIntersectPoint` provided to tools

### DIFF
--- a/src/device/index.js
+++ b/src/device/index.js
@@ -335,7 +335,8 @@ realityEditor.device.postEventIntoIframe = function(event, frameKey, nodeKey) {
                         y: raycastIntersects[0].point.y,
                         z: raycastIntersects[0].point.z,
                         // NOTE: to transform a normal, you must multiply by the transpose of the inverse of the model-view matrix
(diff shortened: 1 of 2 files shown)


dangond-ptc:
Looks good to me! The direction is relative to the world coordinate system, correct?

hobinjk-ptc (approved):

dangond-ptc (approved):

benptc:
Missed a transformation the first time around but its in the world coordinate system now.

There's still an inaccuracy in here, that I'll correct later when I have more time to think it over, where the ray direction drifts away from the pointer as you move away from the origin.

(this screenshot shows it from one perspective – red spheres are translated towards the camera along the ray direction, blue spheres are translated in the opposite direction – they should be perfectly aligned if they're along the correct ray from the camera, which seems to only happen at the origin)

Screen Shot 2022-05-26 at 4 54 05 PM

PR #228, opened 2022-06-13
add screenTopOffset to move envelope X button lower

benptc:
Allows remote operator menubar to adjust the top alignment. Could also be used in future by portrait mode app to avoid the screen notch at the top.


dangond-ptc (approved):
Looks good to me!

hobinjk-ptc (approved):

PR #232, opened 2022-07-05
Ignore camera control clicks; Add offline check

hobinjk-ptc:
Mass-blocks all middle mouse button and right mouse button clicks using a helper function. Adds a button property to the frame's pointer event as well (even though it should only be button: 0 in theory). Adds a notification (shamelessly stolen from the scanning api) about when the network is offline

### DIFF
--- a/src/device/index.js
+++ b/src/device/index.js
@@ -346,6 +346,7 @@ realityEditor.device.postEventIntoIframe = function(event, frameKey, nodeKey) {
             type: event.type,
             pointerId: event.pointerId,
             pointerType: event.pointerType,
+            button: event.button,
             x: newCoords.x,
             y: newCoords.y
         }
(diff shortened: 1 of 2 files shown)


src/device/index.js
```diff
@@ -587,13 +588,25 @@ realityEditor.device.disablePinchToScale = function() {
...
 
+/**
+ * @return {boolean} If the event is intended to control the camera and not the
+ *   AR elements
+ */
+realityEditor.device.isMouseEventCameraControl = function(event) {
+  // If mouse events are enabled ignore right clicks and middle clicks
+  return realityEditor.device.environment.requiresMouseEvents() &&
+    (event.button === 2 || event.button === 1);
+};
```
benptc, line 599: Nice refactor 👍

src/device/onLoad.js
```diff
@@ -277,6 +277,41 @@ realityEditor.device.onload = function () {
         }, 1000);
     }
 
+    fetch(window.location + '/?offlineCheck=' + Date.now()).then(res => {
```
benptc, line 280: Seems like a nice feature!

benptc (approved):
LGTM

PR #233, opened 2022-07-05
uses three-mesh-bvh to speed up raycasts

benptc:
Fixes performance issues when raycasting against large gltf models

### DIFF
--- a/index.html
+++ b/index.html
@@ -74,7 +74,6 @@
     <script src="thirdPartyCode/toolsocket/index.js"></script>
     <script src="thirdPartyCode/hull.js"></script>
     <script src="thirdPartyCode/stats.min.js"></script>
-    <script src="thirdPartyCode/three-mesh-bvh.module.js" type="module"></script>
 
     <!--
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -74,6 +74,7 @@
     <script src="thirdPartyCode/toolsocket/index.js"></script>
     <script src="thirdPartyCode/hull.js"></script>
     <script src="thirdPartyCode/stats.min.js"></script>
+    <script src="thirdPartyCode/three-mesh-bvh.module.js" type="module"></script>
```
hobinjk-ptc, line 77: This line shouldn't be necessary since threejsScene imports three-mesh-bvh as a library. Having the script tag will run the code in three-mesh-bvh as if it were a standalone thing

hobinjk-ptc (changes requested):
It's so great to have this! Looking forwards to seeing the effects it has on the drawing tool especially. One comment inline on some potential redundancy to reduce (unless the file has side effects I'm unaware of)

index.html
benptc: Thanks for the correction!

PR #235, opened 2022-07-07
Move main three.js camera instead of container objects

benptc:
Finally made the change so that the camera moves in the threejsScene, simplifying the coordinate systems.
The threejs mainContainerObj.matrix is set to the groundPlaneSceneNode.worldMatrix, rather than the modelViewMatrix.

Consequences:
- the math for dragging the groundPlaneAnchors was broken, so I took the opportunity to add THREE.TransformControls to translate them in a more standard way
-  @dangond-ptc's areaCreator calculateGroundPlaneIntersection needed an adjustment (but could be simplified a bit – correct me if I'm wrong but it didn't seem necessary to use the getPointAtDistanceFromCamera anymore)
- the main threejs scene lighting was messed up, because it was actually just being lit by a spotlight that happened to cover it because of the way the scene was transformed. I tried replacing it with an ambient light but our area target material doesn't seem to be affected by ambient lighting, so I added a set of directional lights as a temporary replacement. Might need some help improving this if anyone has time to play with it in the future.
- the camera following (in the remote-operator-addon) needed to be updated because of some calculations it did using threejs
- the virtualizers aren't affected at all by the change 👍

### DIFF
--- a/src/gui/ar/groundPlaneAnchors.js
+++ b/src/gui/ar/groundPlaneAnchors.js
@@ -27,6 +27,7 @@ createNameSpace("realityEditor.gui.ar.groundPlaneAnchors");
     let originColor = 0xffffff;
     let mouseCursorMesh = null;
     let initialCalculationMesh = null;
+    let transformControls = {};
 
     function initService() {
         // Note that, currently, positioningMode blocks touch events from reaching anything else, so it should be toggled off when not in use
@@ -99,9 +100,6 @@ createNameSpace("realityEditor.gui.ar.groundPlaneAnchors");
         // set the anchor matrix by taking the x, z position
         knownAnchorNodes[frameKey].setLocalMatrix(anchoredMatrix);
 
-        // we use localMatrix, not world matrix, because mesh is already a child of the ground plane
-        // realityEditor.gui.threejsScene.setMatrixFromArray(threejsGroups[frameKey].matrix, knownAnchorNodes[frameKey].localMatrix);
-        
         threejsGroups[frameKey].position.set(relativeMatrix[12], 0, relativeMatrix[14]);
     }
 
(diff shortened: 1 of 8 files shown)


src/gui/ar/groundPlaneAnchors.js
```diff
@@ -177,26 +136,28 @@ createNameSpace("realityEditor.gui.ar.groundPlaneAnchors");
     // helper function to create the geometry for a surface anchor, including its X-Z axis handles
     function createAnchorGroup(frameKey) {
         const THREE = realityEditor.gui.threejsScene.THREE;
-        const group = new THREE.Group();
+
+        let originSize = 100;
```
hobinjk-ptc, line 140: It would be nice to have this origin box scale the same way as the transform controls (it looks like they're a constant size regardless of distance). Alternatively vice-versa and have the transform controls a fixed size. Totally just an optional polish thing that isn't necessary if it'll be a pain to do

hobinjk-ptc (approved):
Excellent improvements!

src/gui/ar/groundPlaneAnchors.js
benptc: I gave it a shot and it worked on remote operator but for some reason it works differently in the AR devices so I'm going to leave it out for now.

src/gui/ar/groundPlaneAnchors.js
```diff
@@ -177,26 +136,32 @@ createNameSpace("realityEditor.gui.ar.groundPlaneAnchors");
...
-        originBox.add(xBox);
-        // originBox.add(yBox);
-        originBox.add(zBox);
+
+        const options = {
+            size: realityEditor.device.environment.variables.transformControlsSize || 1,
+            hideY: true
+        }
+
+        let transformControl = realityEditor.gui.threejsScene.addTransformControlsTo(group, options, (e) => {
```
dangond-ptc, line 150: nit: this could be simplified to `let transformControl = realityEditor.gui.threejsScene.addTransformControlsTo(group, options, onChange, onDraggingChanged);`

dangond-ptc (approved):
Looks good to me!

PR #242, opened 2022-07-12
Support VAO extension in glproxy

hobinjk-ptc:


dangond-ptc (approved):
I have to admit I don't fully understand how the gl proxy does things, but looks good to me?

hobinjk-ptc:
Thanks! I've added some documentation to this and its corresponding edge-server PR so that the extVao usage is a bit clearer. The tl;dr is that a vertex array object will store all the bindWhatever calls that we (and three.js) would otherwise be redoing every frame and replace them with a bindVertexArray.

The complication is that all the VAO functions are stored on the extension object returned by gl.getExtension which means that we need to mock all the functions on the object in a way that tells the gl proxy on the other side to call the functions on the object instead of on the gl context

PR #243, opened 2022-07-12
Avatar objects + collaborative laser pointers

benptc:
- Automatically creates avatar object when localizes against target, with an invisible tool and node to store the touchState
- Userinterface ignores avatar objects in other aspects of the code (e.g you can't attach tools from pocket to them)
- Writes touch state (and 3d intersection point) into this node's public data
- Subscribes to publicData of other avatar objects and renders a laser beam when they send a touch down state
- renders a small 2d cursor to indicate when you are sending beams

Requires https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/555


src/network/index.js
```diff
@@ -906,8 +909,15 @@ realityEditor.network.onAction = function (action) {
         if (thisAction.reloadObject.object in objects) {
 
             let urlEndpoint = realityEditor.network.getURL(objects[thisAction.reloadObject.object].ip, realityEditor.network.getPort(objects[thisAction.reloadObject.object]), '/object/' + thisAction.reloadObject.object);
+
             this.getData(thisAction.reloadObject.object, thisAction.reloadObject.frame, null, urlEndpoint, function (objectKey, frameKey, nodeKey, res) {
 
+                if (!res) {
+                    delete objects[objectKey];
+                    // TODO: perhaps we need to add onObjectDeleted callbacks that would be triggered here
```
hobinjk-ptc, line 917: For now a console.warn would be good to have here

index.html
```diff
@@ -74,6 +74,7 @@
     <script src="thirdPartyCode/toolsocket/index.js"></script>
     <script src="thirdPartyCode/hull.js"></script>
     <script src="thirdPartyCode/stats.min.js"></script>
+    <script src="thirdPartyCode/three-mesh-bvh.module.js" type="module"></script>
```
hobinjk-ptc, line 77: This can be deleted

hobinjk-ptc (approved):

PR #245, opened 2022-07-12
properly detects world objects saved on 127.0.0.1

benptc:
Pretty sure this is all that's required. I don't have my environment set up to fully test it.


hobinjk-ptc (approved):
LGTM, thanks!

PR #247, opened 2022-07-18
Avatar object 🤝 Network getData API

hobinjk-ptc:
On the proxy we get fancy with 204 status codes. network.getData does not enjoy this


src/avatarObjects.js
```diff
@@ -110,6 +110,9 @@ createNameSpace("realityEditor.avatarObjects");
             if (!cachedWorldObject) {
                 cachedWorldObject = realityEditor.worldObjects.getBestWorldObject();
             }
+            if (!cachedWorldObject) {
```
benptc, line 113: thanks for the edge case improvements

benptc (approved):
LGTM

PR #251, opened 2022-07-21
Ground Plane Visualization Updates

benptc:
Swapped GridHelper to InfiniteGridHelper, set groundplane color on remote operator to grey, and turn on grid by default on remote operator to get a nice baseline visual for the space.

In AR, grid is still cyan, and I added three.js objects to better visualize the origin and where your forward vector hits the groundplane. Toggles off by default on AR devices, since it's more of a debugging tool in these scenarios.

Remote Operator:
Screen Shot 2022-07-21 at 3 33 21 PM

AR Devices (groundplane snaps to new when I press the button on the top right):
groundplane-updates-ios


dangond-ptc (approved):
Looks good to me!

hobinjk-ptc (approved):

PR #255, opened 2022-08-02
Avatar refactor and finalized visual style

benptc:
Stable implementation of avatar objects. Refactored avatarObject.js into avatar/index.js, avatar/draw.js, avatar/network.js, and avatar/utils.js. Adds a user name settings field and renders the user initials at the avatar laser pointer.


src/avatar/draw.js
```diff
@@ -0,0 +1,355 @@
...
+        overlay.style.transform = 'translate3d(' + screenX + 'px, ' + screenY + 'px, 1201px)';
+        overlay.style.display = isVisible ? 'inline' : 'none';
+    }
+
+    // Shows an "Establishing Connection..." --> "Connected!" label in the top left
+    function renderConnectionFeedback(isConnected) {
+        if (!statusUI) {
+            statusUI = document.createElement('div');
+            statusUI.id = 'avatarStatus';
+            statusUI.style.pointerEvents = 'none';
```
hobinjk-ptc, line 284: All of these rules except for the dynamic `top` could be put in a css rule to keep things a bit more cleanly separated

src/avatar/draw.js
```diff
@@ -0,0 +1,355 @@
...
+    function renderConnectionDebugInfo(connectionStatus, debugConnectionStatus, myId, debugMode) {
+        if (!debugMode) {
+            if (debugUI) { debugUI.style.display = 'none'; }
+            return;
+        }
+
+        if (!debugUI) {
+            debugUI = document.createElement('div');
+            debugUI.id = 'avatarConnectionStatus';
+            debugUI.style.pointerEvents = 'none';
```
hobinjk-ptc, line 321: See above, having these in a css rule might even allow us to eventually standardize all the UI zIndices with a shared rule or css variable but that feels a bit far out for now

src/avatar/network.js
```diff
@@ -0,0 +1,186 @@
+createNameSpace("realityEditor.avatar.network");
+
+/**
+ * @fileOverview realityEditor.avatar.network
+ * Contains a variety of helper functions for avatar/index.js to create and discover avatar objects,
+ * realtime broadcast my avatar's state, and subscribe to the state of other avatars
+ */
+
+(function(exports) {
+    const DATA_SEND_FPS_LIMIT = 60;
```
hobinjk-ptc, line 10: This should be lower to play nice with the cloud proxy. Usually I've been going with 10 fps but this might be able to go a bit higher like 30 or so. The only issue with having it be this high is that there's no backpressure on the socket so the server can't say "hey slow this down while I recover"

benptc: I'll try 30 FPS, if you have issues with it feel free to lower it back to 10

PR #259, opened 2022-08-08
adds first version of avatar connected list of icons at the top center

benptc:
Adds icons in the top center of the screen if there are at least two avatars connected, showing how many clients are currently viewing the space. Lots of room for visual improvement, this is just the baseline.


src/avatar/draw.js
```diff
@@ -28,6 +28,59 @@ createNameSpace("realityEditor.avatar.draw");
...
+        }
+
+        // TODO: add initials to each icon that has them
+        // TODO: add hover pointerevents to show full name
+        // TODO: if more than 3-5 icons, show a "... +1" at the end and limit how many icons
+        // TODO: treat your own icon differently
+        let index = 0;
+        for (const [_objectKey, info] of Object.entries(connectedAvatars)) {
+            let initials = realityEditor.avatar.utils.getInitialsFromName(info.name) || '';
+            console.log(initials);
```
hobinjk-ptc, line 67: This would be a great place for the avatar color as well! Also if you want something in the meantime you could use icon.title for hover text

src/avatar/draw.js
```diff
@@ -28,6 +28,59 @@ createNameSpace("realityEditor.avatar.draw");
...
+            
+            let icon = document.createElement('img');
+            icon.classList.add('avatarListIcon', 'avatarListIconVerticalAdjustment');
+            icon.src = '../../../svg/avatar-placeholder-icon.svg';
+            icon.style.left = (40 * index) + 'px';
+            iconContainer.appendChild(icon);
+            
+            index++;
+        }
+        let iconsWidth = (index * 30) + ((index-1) * 10);
```
hobinjk-ptc, line 77: After 40 hours of consultation with a calculator and wolfram alpha I can semi-confidently state that this can be simplified to index * 40 - 10

hobinjk-ptc (approved):
Overall LGTM

src/avatar/draw.js
benptc: Good idea!

src/avatar/draw.js
benptc: Ah yes, that would do it 🔢  thanks!

PR #263, opened 2022-08-12
Misc QC features

benptc:
- Shows UI if we receive a failure message about camera permissions when trying to start vuforia
- Adds new envelope X button that matches new visual style

- Various changes that enable new code I wrote in the pop-up-addon
  - Adds pause/resumeObjectDetections APIs that the pop-up-app can use when in the menu to avoid loading old scans, but allowing a specific object name (your new scan) to be added as an exception so that scanning still works
    - I'm planning to refactor this nicely after this PR, just wanted to get a first version in, in case we QC it again on Mon
  - Triggers a callback when Vuforia successfully adds a tracking dataset, which the pop-up-app can use to tell you when to start looking for your area target (after the screenshot is uploaded too, not when the scan finishes generating)

camera-access-prompt copy


src/app/callbacks.js
```diff
@@ -74,7 +74,24 @@ createNameSpace('realityEditor.app.callbacks');
...
+            while (listeners.onVuforiaInitFailure.length > 0) { // dismiss the intializing pop-up that was waiting
+                let callback = listeners.onVuforiaInitFailure.pop();
+                callback();
+            }
+            
+            let headerText = 'Needs camera access';
+            let descriptionText = 'Please enable camera access<br/>in your device\'s Settings app,<br/>and try again.';
+
+            let notification = realityEditor.gui.modal.showSimpleNotification(
+                headerText, descriptionText,function () {
```
hobinjk-ptc, line 89: nit: space after comma

hobinjk-ptc (approved):
Great improvements to have! Especially excited for the object tracking pause workaround

PR #265, opened 2022-08-16
Misc improvements

hobinjk-ptc:


benptc (approved):
thanks for this, too! :)

PR #266, opened 2022-08-17
Glproxy fallback restore

hobinjk-ptc:
Prevents frames from flickering out of existence during difficult moments. Fixes issue where opening the spatial draw tool would cause the camera-vis playback tool to change its winding order or something


dangond-ptc (approved):
Looks good to me!

PR #267, opened 2022-08-17
Only handle every third touchmove when mouse isn't pressed

hobinjk-ptc:


dangond-ptc:
Would it be better to have it be a timeout-based system (e.g. if it has been less than 100ms since last touchmove, ignore)? Especially since the events are framerate independent and can fire very quickly

hobinjk-ptc:
Yeah, I agree with that approach being better

dangond-ptc (approved):
Looks good to me!

PR #273, opened 2022-08-30
add scaffolding for humanPose objects

benptc:


hobinjk-ptc (approved):
\o/ let's get this bread

benptc:
The old human pose implementation exists essentially entirely in the server (e.g. https://github.com/ptcrealitylab/vuforia-spatial-edge-server/blob/b13ea4aa819cb44fee19011e79a327fda1d36e9a/server.js#L3828) – I can update/remove that later as it won't conflict with this version as long as we continue not to invoke it.

PR #274, opened 2022-09-01
HumanPoseObject creation from received poses

benptc:
Creates a humanPose object and sets the localMatrix of each of its frames based on the received poses. Updates the sceneGraph/network.js to automatically write frames' positions (persistently) to the server every X seconds if they've changed, similar to how it uploads the object positions.

Requires https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/588


hobinjk-ptc (approved):
Excellent!

PR #275, opened 2022-09-02
rebaScore support

hobinjk-ptc:
Notably changes JOINT_CONNECTIONS to be a bit more organization-friendly


benptc (approved):
Good update for JOINT_CONNECTIONS, and nice to have the rebaScore script in the userinterface codebase now!

PR #276, opened 2022-09-09
Fix merged raycast; Improve avatar pointermove filtering

hobinjk-ptc:
Before:
before

After:
after-both

I believe this closes the book on Vuforia 10.9 making area targets multi-mesh mapping mayhem


dangond-ptc (approved):
Looks good to me!

PR #285, opened 2022-10-07
Wrap native APIs in Promises for cleaner asynchronous getters

benptc:
While implementing the `doesDeviceHaveDepthSensor` API, it felt about time to make a cleaner way to call them. Tried not to refactor too many APIs at once, but wrapped a few more in this style for now.

### DIFF
--- a/src/app/promises.js
+++ b/src/app/promises.js
@@ -1,0 +1,67 @@
+createNameSpace("realityEditor.app.promises");
+
+// provides a simpler interface to the native APIs
+(function(exports) {
+    // assume APIs fail if they take longer than 3 seconds to resolve, they should be almost instantaneous
+    const TIMEOUT = 3000;
+    const app = realityEditor.app;
+
+    // promises must be resolved externally due to the inner workings of how callbacks signatures get passed to swift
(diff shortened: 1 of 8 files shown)


src/app/promises.js
```diff
@@ -0,0 +1,76 @@
...
+    // exposes randomly generated public function signatures to resolve the deferred promises when the native code returns
+    exports._callbackProxies = {};
+
+    // Helper function to wrap the appFunctionCall in a deferred promise that will resolve when the native code finishes
+    // The name of each resolve param should be included iff the native API returns multiple values
+    function makeAPI(appFunctionCall, resolveParams) {
+        return function() {
+            // when the API is called, create a new Promise
+            let deferred = new Deferred(() => {
+                delete realityEditor.app.promises._callbackProxies[functionUuid];
```
hobinjk-ptc, line 48: nit: placing this here makes me wonder where functionUuid comes from even though it's technically valid since the declaration will be hoisted

hobinjk-ptc (approved):
Excellent! Really great to have a modern API start to seep in

PR #289, opened 2022-10-12
update target downloader to use promise version of addNewMarkerJPG

benptc:
Error message was occurring if the old version of addNewMarkerJPG was called, since it was referring to the old callback signature which is no longer public.


hobinjk-ptc (approved):
Excellent! LGTM

PR #290, opened 2022-10-17
Render poses+REBA directly from Object data structure

benptc:
Update HumanPoseRenderer to properly render the pose joint tools.

Works in AR and on remote operator (can comment out `this.drawSkels(skels);` in remote-operator-addon RealityZoneViewer.js to test without rendering double)


hobinjk-ptc (approved):
Awesome! So cool to see the 3d poses rendering on the phone

PR #292, opened 2022-10-19
Fix alwaysFaceCamera to use proper lookAt calculations

benptc:
Adds `sceneGraph.getModelViewMatrixLookingAt(id, idToLookAt)` function to properly compute the modelView matrix for any sceneNode such that it is rotated towards any other sceneNode.

Uses it in `alwaysFaceCamera` API to make tools properly look at the camera.


hobinjk-ptc (approved):
Sure is math!

PR #293, opened 2022-10-20
Human Pose Objects

benptc:
This has all been code-reviewed along the way, but adding you as reviewers to communicate following implications:

1) Remote operator will now have duplicate pose-rendering code that should probably be removed

2) AR app will now render the 2D poses and the 3D poses in an overlapping manner. I have discussed today with Martin the possibility of keeping the 2D pose rendering as a toggle-able option in the app, to help with debugging the raw vs processed poses, but that toggle isn't included yet.

### DIFF
--- a/index.html
+++ b/index.html
@@ -104,7 +104,8 @@
 
     <script src="src/humanPose/index.js"></script>
     <script src="src/humanPose/network.js"></script>
-    <script src="src/humanPose/draw.js"></script>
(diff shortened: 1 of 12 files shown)


hobinjk-ptc (approved):
I'll work on removing the human pose rendering code from the remote operator and hooking up the menu shortcuts to this new approach

dangond-ptc (approved):
Lots of changes but everything looks good to me!

PR #295, opened 2022-10-20
fixes trashcan so tool gets deleted even if it drifts away from the pointer

benptc:
Moves deletion check to the document pointerup event instead of the iframe pointerup event, so it triggers as long as you lift your mouse/finger in the designated region (the right edge of the screen), even if the iframe drifted away from your pointer.

In future I also hope to improve the dragging calculations so that the tools don't drift away from the pointer in the first place, but this should improve robustness either way.


hobinjk-ptc (approved):
Excellent fix! I've run into this one oh so many times

PR #298, opened 2022-10-24
Add tools at the spatial cursor, and snap cursor to overlapping tools

benptc:
1. This adds the ability to add tools at the spatial cursor position. Extra calculations are included to rotate the added tools so that they appear upright relative to the global up vector.
2. I've also added a new helper function, `realityEditor.sceneGraph.convertToNewCoordSystem(point, oldSystem, newSystem)` to help simplify calculations in the future.
3. I also added some functions to the spatialCursor/index.js to snap the position and rotation of the cursor to the overlapping tool iframe.

### DIFF
--- a/src/spatialCursor/index.js
+++ b/src/spatialCursor/index.js
@@ -176,12 +176,14 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         return worldIntersectPoint; // these are relative to the world object
     }
 
+    // gets the position relative to groundplane (common coord system for threejsScene)
     function getToolPosition(toolId) {
         let toolSceneNode = realityEditor.sceneGraph.getSceneNodeById(toolId);
         let groundPlaneNode = realityEditor.sceneGraph.getGroundPlaneNode();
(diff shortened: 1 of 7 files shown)


src/spatialCursor/index.js
```diff
@@ -170,5 +176,81 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         return worldIntersectPoint; // these are relative to the world object
     }
 
+    function getToolPosition(toolId) {
+        let toolSceneNode = realityEditor.sceneGraph.getSceneNodeById(toolId);
+        let groundPlaneNode = realityEditor.sceneGraph.getGroundPlaneNode();
+        return realityEditor.sceneGraph.convertToNewCoordSystem({x: 0, y: 0, z: 0}, toolSceneNode, groundPlaneNode);
+    }
+
+    function getToolForwardVector(toolId) {
```
hobinjk-ptc, line 185: It would be nice to have an explanation of what "forward" means in this context

hobinjk-ptc (approved):
Excellent! Very excited to have this feature

PR #299, opened 2022-10-26
Disable repeated tapping when no frames visible

hobinjk-ptc:


benptc (approved):
Thanks for taking care of this one!

PR #304, opened 2022-11-01
Tool dragging improvements

benptc:
PR #302, which improved the dragging of 3D tools, had a couple issues:
1) when physically moving the phone around while holding a tool (in "unconstrained editing mode") the new repositioning method would fight with the old unconstrained method and the tool would flicker.
2) it was still using old calculations for moving 2D tools along their local XY plane, which was using some hacky browser/CSS tricks to do the calculations

This PR adds the following:
1) "unconstrained editing" simply makes the tool a child of the camera instead of freezing its relative matrix, which causes it to move with the camera without fighting with new local matrix values provided by the dragging mechanism
2) new `realityEditor.gui.ar.positioning.moveVehicleAlongPlane` function uses more elegant ray-plane intersection function to calculate dragging the 2D tools along their XY plane instead of webkit/CSS-dependent hacks
3) some (messier than ideal) adjustments in order to still realtime sync and upload the position of the tool relative to the object even if its sceneNode is parented to the camera

new-2d-repositioning

new-3d-repositioning

### DIFF
--- a/src/device/index.js
+++ b/src/device/index.js
@@ -411,14 +411,20 @@ realityEditor.device.resetEditingState = function() {
     // gets triggered before state gets reset, so that subscribed modules can respond based on what is about to be reset
     this.callbackHandler.triggerCallbacks('resetEditingState');
     
+    // properly write the vehicle position to the server if it's been moved relative to another parent
     if (this.getEditingVehicle() && this.isEditingUnconstrained(this.getEditingVehicle())) {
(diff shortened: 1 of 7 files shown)


hobinjk-ptc (approved):
Excellent! And a nice simplification to tear out the old code

dangond-ptc (approved):
Looks good to me! Thanks for adding this!

PR #306, opened 2022-11-04
adds a flag that remote operator can use to disable the creation of memories

benptc:
(memories don't really work anymore but creation is currently available by double clicking the background).
We could bring it back with a little bit of work, but pop-up-addon can disable it too for now, since it doesn't include the memory pocket in the new UI.


hobinjk-ptc (approved):
😢

benptc:
> 😢

I know :( definitely worth restoring in future!

PR #309, opened 2022-11-07
disable spatial cursor computations while its toggled off; temporarily remove cursor snapping

benptc:
The `getAllDivsUnderCoordinate` function that the cursor uses was breaking pocket scroll and settings menu, and hurting performance – sets display:none and then resets display:inline to all overlapping elements during the calculation. I'm going to update this function to use pointer-events rather than display in the future, but disabling it for now as a temporary fix since the pointer-events way has some complications with conflicting CSS definitions.


hobinjk-ptc (approved):
I have pushed a commit :)

hobinjk-ptc:
Thanks for finding the root cause! Consider the commit payment

benptc:
Ah, thanks for the commit! Should have consulted the web APIs 😅

PR #313, opened 2022-11-10
Human pose refactor

hobinjk-ptc:
Highly recommend viewing this with ?w=1 on the diff url

- Refactors out analysis-related functions in draw.js to be a modular class that can be ripped out later
- Rewrites every humanPose/* file as an es6 module that is re-exported by humanPose/index.js as realityEditor.humanPose.*

This rewrite and re-export is a pretty good model for how we can migrate the entire user interface to es6 modules which will in turn enable load time improvements through better tooling


dangond-ptc (approved):
Very nice to have more and more things available as imports!

hobinjk-ptc:
When everything's an import and we can flip the switch to bring the user interface from 150 script requests to 1 it's going to be so good

src/humanPose/draw.js
```diff
@@ -1,430 +1,456 @@
...
+/**
+ * Renders COCO-pose keypoints
+ */
+class HumanPoseRenderer {
+    /**
+     * @param {string} id - Unique identifier of human pose being rendered
+     * @param {THREE.Object3D} historyLineContainer - THREE container for
+     *                         history line meshes
+     * @param {THREE.Object3D} historyCloneContainer - THREE container for
+     *                         history clone meshes
```
benptc, line 19: nit: looks like the jsdoc params can be removed

benptc (approved):
Great to see a template for refactoring, plus the updates to humanPose/draw!

PR #315, opened 2022-11-28
Cleanup of various remote operator loading race conditions

hobinjk-ptc:
Fixes an intermittent bug on Safari/Firefox where the remote operator's content scripts could be loaded later than various initService calls that relied on the remote operator's environment overrides. This directly overrides ar tracking to prevent a flash of the popup. It also adjusts the logic of desktopSocket creation to fix a race condition in a way that may be more universally applicable


benptc (approved):
Looks good! Thanks for cleaning up this logic

PR #316, opened 2022-11-28
Use dataset.didNotAcceptTouch and getAllDivsUnderCoordinate in touch propagation

hobinjk-ptc:
This new approach cleanly cycles through candidate touch elements
without style modification (setting display: none). This prevents a
potential flash of elements hiding/unhiding themselves during the touch
propagation cycle.

Very high possibility for subtle bugs considering how fundamental this code is


hobinjk-ptc:
Speak of the devil, as I made this PR the weird bug I encountered at the very start of this rewrite came back so I'm marking this as a draft until I get more time to work on it

benptc:
> Speak of the devil, as I made this PR the weird bug I encountered at the very start of this rewrite came back so I'm marking this as a draft until I get more time to work on it

Sounds good. I think I tried improving this in the past and also ran into bugs. Would be a great improvement if we ensure it works.

PR #317, opened 2022-11-30
MeshPath and SpaghettiMeshPath

benptc:
MeshPath class is added that can often be used to replace MeshLine with an extruded rectangular line of variable width and color.

SpaghettiMeshPath extends MeshPath with user interaction, including the ability to measure distance and time between two points on the path.

I haven't actually updated the humanPose history lines to use this yet – first use will come in the remote-operator addon where I've updated the CameraVis history lines to use it, as shown here:

```
this.historyPoints = [];
this.historyMesh = new SpaghettiMeshPath(this.historyPoints, {
    width_mm: 30,
    height_mm: 30,
    usePerVertexColors: true,
    colorBlending: false,
    wallBrightness: 0.6
});
```

spaghettiMeshPath-demo


src/sceneGraph/index.js
```diff
@@ -386,8 +389,11 @@ createNameSpace("realityEditor.sceneGraph");
         }
     }
 
-    function getWorldPosition(activeKey) {
-        let sceneNode = getSceneNodeById(activeKey);
+    function getWorldPosition(idOrNode) {
```
hobinjk-ptc, line 392: Taking either a string or a scenenode has the potential to cause this function to not be optimized as much as if it were split into getWorldPositionById(id) which calls getWorldPosition(sceneNode) or equivalent

src/sceneGraph/index.js
```diff
@@ -744,15 +750,7 @@ createNameSpace("realityEditor.sceneGraph");
...
-            rootNode.updateWorldMatrix();
-        }
-        if (cameraNode) {
-            cameraNode.updateWorldMatrix();
-        }
-
-        // this might become a child of a world object but still safe to call update multiple times
-        if (groundPlaneNode) {
-            groundPlaneNode.updateWorldMatrix();
+            rootNode.updateWorldMatrix(); // updates everything that is a child of the root, too (objects, camera, groundPlane, etc)
```
hobinjk-ptc, line 753: Nice simplification!

dangond-ptc (approved):
Looks good to me!

src/sceneGraph/index.js
benptc: Good point. I'll revert to previous implementation for now and defer that refactor for later.

hobinjk-ptc (approved):
Very excited to test this out!

PR #319, opened 2022-12-08
Area target frustum culling shader

benptc:
Instantiates a ViewFrustum for each virtualizer that sends an update message to the threejsInterface. Discards points that fall within the volume of any of the frustums. Supports up to MAX_VIEW_FRUSTUMS clients at a time, which I currently set to 5.


hobinjk-ptc (approved):
Excellent work! Tested locally and LGTM

PR #320, opened 2022-12-09
Add layers and internals retrieval to threejs scene

hobinjk-ptc:


hobinjk-ptc:
I've adjusted this to stay with the conventional rendering order since the frustum shader exists now but something I was doing and could put back in was drawing the gltf model below everything else in the scene. Thoughts?

benptc (approved):
I think this could still be useful! I'm experimenting now with minimizing the visual effect of the "shadows" cast by the frustums and one potential solution might involve rendering the wireframe mesh of the scene underneath everything

hobinjk-ptc:
Updated accordingly! You can see the required dance in the changes

PR #321, opened 2022-12-12
Added joint visibility visualisation to 2D skeleton rendering.

ptc-mklaudiny:


src/humanPose/draw.js
```diff
@@ -321,6 +321,10 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
 
 
     function renderHumanPoseObjects(poseObjects) {
+
+        // MK HACK: to prevent rendering on mobile 
+        //if (!realityEditor.device.environment.requiresMouseEvents()) return;
```
hobinjk-ptc, line 326: You could replace this with a flag where if 2d pose drawing is enabled then 3d pose drawing is disabled

ptc-mklaudiny: done

PR #322, opened 2022-12-12
Human pose history spaghetti

hobinjk-ptc:
- [x] memory management
- [x] Improve various spaghetti interactions
- [x] Make sure that live spaghetti also works
- [x] refactor code to be less gros (and gross)


dangond-ptc (approved):
Looks good to me!

PR #324, opened 2022-12-14
Increase update framerate to match pose frequency

hobinjk-ptc:


dangond-ptc (approved):
LGTM, if only improving framerate were always so easy

PR #325, opened 2022-12-14
Add guard to use of poseObject.matrix

hobinjk-ptc:


ptc-mklaudiny:
Great that you tracked it down. I suppose that the .matrix member is added first time in HumanPoseObject.prototype.updateJointPositions() Should it be maybe part of object constructor?

src/humanPose/draw.js
```diff
@@ -643,9 +643,11 @@ function updateJointsHistorical(poseRenderer, poseObject) {
 function updateJoints(poseRenderer, poseObject) {
     let groundPlaneRelativeMatrix = getGroundPlaneRelativeMatrix();
 
-    let objectRootMatrix = new THREE.Matrix4();
-    setMatrixFromArray(objectRootMatrix, poseObject.matrix);
-    groundPlaneRelativeMatrix.multiply(objectRootMatrix);
+    if (poseObject.matrix && poseObject.matrix.length > 0) {
```
ptc-mklaudiny, line 646: Should we even render if this matrix is not in valid state yet?

src/humanPose/draw.js
```diff
@@ -643,9 +643,11 @@ function updateJointsHistorical(poseRenderer, poseObject) {
...
+        let objectRootMatrix = new THREE.Matrix4();
+        setMatrixFromArray(objectRootMatrix, poseObject.matrix);
+        groundPlaneRelativeMatrix.multiply(objectRootMatrix);
+    }
```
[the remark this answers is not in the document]
hobinjk-ptc: Yes, each frame is relative to its parent object and objectRootMatrix is (in theory) the parent object's matrix. As far as I can tell this matrix will never be initialized on the phone ui, causing this bug.

src/humanPose/draw.js
hobinjk-ptc: On the phone the matrix will never be updated to a valid matrix

hobinjk-ptc:
I've pushed a commit removing all the uses of `[]` as a placeholder for the identity matrix. This adjustment on its own would probably fix this bug (the phone not repositioning the pose's containing parent object) but I like defense in depth

ptc-mklaudiny (approved):
thanks for clarifications and making it more robust.

PR #328, opened 2023-01-11
Fade frustum cutout based on viewing angle; prepare spatial cursor for new features

benptc:
One last update for the area target shader: based on the similarity between the viewing angle and the virtualizer angle, the mesh wireframe fades in and out. Since you mostly want to see the shadows filled in when you're looking at it from different angles. And when you're perfectly aligned with the virtualizer, you only see the live image data. This minimizes distracting visual effects while you're following a virtualizer.

This PR also includes an update for the spatial cursor to make it follow the remote operator mouse, and to allow it to instantiate new tools at its location. But those effects are toggled off now with two flags: DEFAULT_SPATIAL_CURSOR_ON=false and ADD_SEARCH_TOOL_WITH_CURSOR=false

wireframe-mesh-with-pointcloud-v6 2023-01-11 16_35_29

### DIFF
--- a/src/spatialCursor/index.js
+++ b/src/spatialCursor/index.js
@@ -5,8 +5,9 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
 (function(exports) {
 
     const SNAP_CURSOR_TO_TOOLS = true;
+    const DEFAULT_SPATIAL_CURSOR_ON = false;
 
-    let isCursorEnabled = true;
+    let isCursorEnabled = DEFAULT_SPATIAL_CURSOR_ON;
     let isUpdateLoopRunning = false;
 
     let cachedWorldObject;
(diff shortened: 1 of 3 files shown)


src/spatialCursor/index.js
```diff
@@ -92,10 +93,53 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
...
+                // raycast against the spatial cursor
+                let intersects = realityEditor.gui.threejsScene.getRaycastIntersects(e.clientX, e.clientY, [indicator2]);
+                if (intersects.length > 0) {
+                    addToolAtScreenCenter('searchDigitalThread');
+                }
+            });
+        }
+    }
+
+    // copied from pop-up-onboarding addon – should we refactor somewhere convenient to both?
```
hobinjk-ptc, line 118: It looks like realityEditor.spatialCursor.addToolAtScreenCenter would be a good place

hobinjk-ptc (approved):

PR #332, opened 2023-01-25
Spatial cursor indicator arrow matching avatar color

Steve-KX-RL:
1. Add support for spatial indicators & off-screen arrows.
2. Update spatial cursor, add mobile & desktop remote operator support
3. Add helper functions in threejsScene.js and avatar/index.js, to use in spatial indicator and off-screen arrow
4. Spatial cursor, indicator, and off-screen arrow all match avatar colors. When can't find an avatar, default to cyan color.

### DIFF
--- a/src/gui/spatialArrow.js
+++ b/src/gui/spatialArrow.js
@@ -31,6 +31,7 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         canvasContainer.style.position = 'absolute';
         canvasContainer.style.top = '0';
         canvasContainer.style.left = '0';
+        canvasContainer.style.pointerEvents = 'none';
         document.body.appendChild(canvasContainer);
 
         canvas = document.createElement('canvas');
--- a/src/gui/spatialArrow.js
+++ b/src/gui/spatialArrow.js
@@ -126,7 +126,7 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
     
     let finalPosX = 0, finalPosY = 0;
     
-    function determineObjectScreenPosition() {
+    function drawArrowsAtIndicatorScreenPositions() {
         
         let worldPos = new THREE.Vector3();
         let screenX, screenY;
@@ -209,9 +209,7 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
     
     function drawIndicatorArrows() {
         searchForIndicators();
-        determineObjectScreenPosition();
-        // draw a single test arrow at the screen center
-        
+        drawArrowsAtIndicatorScreenPositions();
     }
 
     function update() {
--- a/src/gui/spatialArrow.js
+++ b/src/gui/spatialArrow.js
@@ -9,8 +9,6 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
     let ctx;
     let screenW, screenH;
     let screenRatio;
-    
-    let color, colorLighter;
 
     function initService() {
         console.log('%c spatial arrow init!', 'color: green');
@@ -20,17 +18,13 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         update();
     }
 
-    // for testing purpose on js fiddle, don't include it in final code
-    const windowRatio = 1;
-
     window.addEventListener('resize', () => {
         resizeCanvas();
         initCanvas();
         update();
     });
-
-    // todo: dynamically set style.top based on menu bar's height -- making the arrow canvas right below the menu bar
-    const menuBarHeight = 30;
+    
+    let menuBarHeight;
 
     function addCanvas() {
         canvasContainer = document.createElement('div');
@@ -43,10 +37,10 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         canvas = document.createElement('canvas');
         canvas.className = 'arrow-canvas';
         canvas.style.position = 'absolute';
+        menuBarHeight = realityEditor.device.environment.variables.screenTopOffset;
         canvas.style.top = `${menuBarHeight}px`;
         canvas.style.left = '0';
         canvas.style.zIndex = '3001';
-        canvas.style.border = '2px solid black';
         canvasContainer.appendChild(canvas);
 
         ctx = canvas.getContext("2d");
@@ -54,17 +48,12 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
 
     function resizeCanvas() {
         if (canvas !== undefined) {
-            screenW = window.innerWidth * windowRatio;
-            screenH = (window.innerHeight - menuBarHeight) * windowRatio;
+            screenW = window.innerWidth;
+            screenH = window.innerHeight - menuBarHeight;
             screenRatio = screenH / screenW;
             canvas.width = screenW;
             canvas.height = screenH;
         }
-    }
-
-    function Rot(a) {
-        let s = sin(a), c = cos(a);
-        return [[c, -s], [s, c]];
     }
 
     const Clamp = (x, low, high) => {
@@ -87,32 +76,16 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         ctx.translate(x, y);
     }
 
-    let rotation = 0;
-
     function Rotate(a) {
-        rotation += a;
         ctx.rotate(a);
     }
 
-    let scaleX = 1, scaleY = 1;
-
     function Scale(x, y) {
-        scaleX *= x;
-        scaleY *= y;
         ctx.scale(x, y);
     }
 
     function clear() {
         ctx.clearRect(-translateX, -translateY, screenW, screenH);
-    }
-
-    // draw a rect at (x, y) center
-    function Rect(x, y, width, height, rotation = 0) {
-        Translate(x, y);
-        Rotate(rotation);
-        ctx.fillRect(-width / 2,-height / 2, width, height);
-        Rotate(-rotation);
-        Translate(-x, -y);
     }
 
     // draw an arrow at (x, y) center
@@ -125,16 +98,13 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         let region = new Path2D();
         region.moveTo(0, -3);
         region.lineTo(-2, 2);
-        // region.lineTo(0, 1);
         region.lineTo(2, 2);
         region.closePath();
         // fill path
         ctx.fillStyle = innerColor;
-        // ctx.fillStyle = innerColor;
         ctx.fill(region, 'evenodd');
         // stroke path
         ctx.strokeStyle = outerColor;
-        // ctx.strokeStyle = outerColor;
         ctx.lineWidth = .4;
         ctx.stroke(region);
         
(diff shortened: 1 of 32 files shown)


src/gui/spatialArrow.js
```diff
@@ -0,0 +1,324 @@
...
+
+(function (exports) {
+    
+    let canvasContainer;
+    let canvas;
+    let ctx;
+    let screenW, screenH;
+    let screenRatio;
+    
+    let color, colorLighter;
```
benptc, line 13: If you run `npm test` or `npm run lint`, you'll see a list of automatically detected code linting errors (you may need to install eslint first) – such as removing variables that are defined but not used. GitHub also highlights these issues when you make the pull request. Can you go through and correct each of these linting errors?

src/gui/spatialArrow.js
```diff
@@ -0,0 +1,324 @@
...
+    function initService() {
+        console.log('%c spatial arrow init!', 'color: green');
+        addCanvas();
+        resizeCanvas();
+        initCanvas();
+        update();
+    }
+
+    // for testing purpose on js fiddle, don't include it in final code
+    const windowRatio = 1;
```
benptc, line 24: should this be removed?

src/gui/spatialArrow.js
```diff
@@ -0,0 +1,324 @@
...
+
+    window.addEventListener('resize', () => {
+        resizeCanvas();
+        initCanvas();
+        update();
+    });
+
+    // todo: dynamically set style.top based on menu bar's height -- making the arrow canvas right below the menu bar
+    const menuBarHeight = 30;
```
benptc, line 33: You can use `realityEditor.device.environment.variables.screenTopOffset` to get this dynamically

src/gui/spatialArrow.js
```diff
@@ -0,0 +1,324 @@
...
+        canvasContainer.style.left = '0';
+        document.body.appendChild(canvasContainer);
+
+        canvas = document.createElement('canvas');
+        canvas.className = 'arrow-canvas';
+        canvas.style.position = 'absolute';
+        canvas.style.top = `${menuBarHeight}px`;
+        canvas.style.left = '0';
+        canvas.style.zIndex = '3001';
+        canvas.style.border = '2px solid black';
```
benptc, line 49: Is there a reason to include the black border around the edge of the screen, or was that just for debugging?

src/gui/spatialArrow.js
```diff
@@ -0,0 +1,324 @@
...
+            screenW = window.innerWidth * windowRatio;
+            screenH = (window.innerHeight - menuBarHeight) * windowRatio;
+            screenRatio = screenH / screenW;
+            canvas.width = screenW;
+            canvas.height = screenH;
+        }
+    }
+
+    function Rot(a) {
+        let s = sin(a), c = cos(a);
```
benptc, line 66: I think this would need to be Math.sin and Math.cos, but since `Rot` isn't used it should also be removed. Alternatively, if you have unused variables or functions that you want to keep, you can name them with an underscore as the first character of the name, to tell the linting system to ignore it. For example, `_Rot`

src/gui/spatialArrow.js
```diff
@@ -0,0 +1,324 @@
...
+        // draw path
+        let region = new Path2D();
+        region.moveTo(0, -3);
+        region.lineTo(-2, 2);
+        // region.lineTo(0, 1);
+        region.lineTo(2, 2);
+        region.closePath();
+        // fill path
+        ctx.fillStyle = innerColor;
+        // ctx.fillStyle = innerColor;
```
benptc, line 133: If you remove some of the commented-out code, that will make it more easily readable

src/gui/spatialArrow.js
```diff
@@ -0,0 +1,324 @@
...
+                // }
+                // // out of bounds
+                // if (screenX > desX) {
+                //     finalPosX = desX;
+                //     finalPosY = desX * m;
+                // } else if (screenX < desX) {
+                //     finalPosX = -desX;
+                //     finalPosY = -desX * m;
+                // }
+                // Arrow(finalPosX, finalPosY, angle, 5);
```
benptc, line 237: Can remove this commented code too

src/gui/spatialIndicator.js
```diff
@@ -0,0 +1,419 @@
...
+    varying vec2 vUv;
+    
+    void main() {
+        vUv = uv;
+        vec4 mvPosition = modelViewMatrix * vec4(position, 1.0);
+        gl_Position = projectionMatrix * mvPosition;
+    }
+    `;
+    
+    const cylinderFragmentShader = `
```
benptc, line 20: Nice shaders! 👍

src/gui/spatialIndicator.js
```diff
@@ -0,0 +1,419 @@
...
+        return Math.min(Math.max(x, low), high);
+    }
+
+    const Remap01 = (x, low, high) => {
+        return Clamp((x - low) / (high - low), 0, 1);
+    }
+
+    const Remap = (x, low1, high1, low2, high2) => {
+        return low2 + (high2 - low2) * Remap01(x, low1, high1);
+    }
```
benptc, line 166: For now it's ok to have these helper functions defined twice (once in spatialArrow, once in spatialIndicator), but if you find yourself needing them in a third place, I'd recommend we refactor them to a utilities file and import them into both.

src/gui/spatialIndicator.js
```diff
@@ -0,0 +1,419 @@
...
+
+    window.addEventListener('pointerdown', (e) => {
+        if (e.button === 0) handleMouseClick(e);
```
benptc, line 169: You might be able to use `!realityEditor.device.isMouseEventCameraControl(e)` instead of `e.button === 0`, to make more consistent with other clicking logic, unless you have a specific reason for checking this way instead

src/gui/spatialIndicator.js
```diff
@@ -0,0 +1,419 @@
...
+                raycastIntersects[0].point.applyMatrix4(inverseGroundPlaneMatrix);
+                let trInvGroundPlaneMat = inverseGroundPlaneMatrix.clone().transpose();
+                worldIntersectPoint = {
+                    point: raycastIntersects[0].point,
+                    normalVector: raycastIntersects[0].face.normal.clone().applyMatrix4(trInvGroundPlaneMat).normalize(),
+                }
+            }
+        }
+        return worldIntersectPoint; // these are relative to the world object
+    }
```
benptc, line 234: I was thinking it's probably better to export this function at the bottom of spatialCursor.js rather than redefine it here. You can give the function there an optional third parameter that lets you customize which objectsToCheck. But I see you're doing some custom logic on the result based on what it hits, so refactoring would become more complex.

benptc (changes requested):
I tested it out and it looks very nice! There's just a number of places in the code that can be simplified, and some code linting errors to fix, before we can merge it. Thanks!

src/gui/spatialArrow.js
Steve-KX-RL: Thx for the tip! I executed npm run lint and removed some unused code. Thx Ben!

src/gui/spatialArrow.js
Steve-KX-RL: Yes. I used another variable screenBorderFactor to calculate a small offset for the arrows to be drawn inside the screen. Forgot to remove this one.

src/gui/spatialArrow.js
Steve-KX-RL: Got it, thx!

src/gui/spatialArrow.js
Steve-KX-RL: Yea this is just for debugging. Just removed it.

src/gui/spatialArrow.js
Steve-KX-RL: Nice tip! I'll keep it in mind. But this is unused so I removed it.

src/gui/spatialArrow.js
Steve-KX-RL: Yes. Just removed all the unused code in spatialArrow.js

src/gui/spatialIndicator.js
Steve-KX-RL: Yea I do think that we need a math utils function file to store all our math functions. VirtualCamera.js also has a lot of math functions that we might need in other files.

src/spatialCursor/index.js
```diff
@@ -106,15 +182,31 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
...
+                    worldIntersectPoint = getRaycastCoordinates(screenX, screenY);
+                    // console.log(worldIntersectPoint.point.x, worldIntersectPoint.point.y, worldIntersectPoint.point.z)
+                    updateSpatialCursor();
+                    updateTestSpatialCursor();
+                    uniforms['time'].value = clock.getElapsedTime() * 10;
+
+                    if (SNAP_CURSOR_TO_TOOLS) {
+                        trySnappingCursorToTools();
+                    }
+                }
```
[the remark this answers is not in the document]
Steve-KX-RL: Good catch. This new logic might have bugs on iPhone b/c setting screenX & screenY in the first "if", they might get accidentally resetted in the "pointerdown" event. So I'll add an "else" after the first "if". Other than that, great!

src/gui/spatialArrow.js
```diff
@@ -0,0 +1,224 @@
...
+        if (canvas !== undefined) {
+            screenW = window.innerWidth;
+            screenH = window.innerHeight - menuBarHeight;
+            screenRatio = screenH / screenW;
+            canvas.width = screenW;
+            canvas.height = screenH;
+        }
+    }
+
+    const Clamp = (x, low, high) => {
```
hobinjk-ptc, line 59: It's more standard to give functions a lowercase name so they're not confused with classes

src/gui/spatialArrow.js
```diff
@@ -0,0 +1,224 @@
...
+    let indicators = [];
+    function searchForIndicators() {
+        // todo: auto detect the indicator names, instead of hard-coded 'cylinderIndicator'
+        indicators = realityEditor.gui.threejsScene.getObjectsByName('cylinderIndicator');
+    }
+    
+    let worldPos = new THREE.Vector3();
+    let screenX, screenY;
+    let screenBorderFactor = 0.97;
+    let desX = 0, desY = 0;
```
hobinjk-ptc, line 131: These variables don't need to be global. It's good practice to declare them closer to their point of use when you can

src/gui/spatialIndicator.js
```diff
@@ -0,0 +1,370 @@
...
+            // col = mix(col, red, boundary);
+        
+            // float boundary = S(0.2, 0.0, 1. - uv.y);
+            // alpha = mix(alpha, 0., boundary);
+        
+            gl_FragColor = vec4(col, alpha);
+        }
+    `;
+
+    const amount = 5;
```
hobinjk-ptc, line 104: Could use a comment here corresponding to the one above mentioning how they need to be kept in sync

src/gui/spatialIndicator.js
```diff
@@ -0,0 +1,370 @@
...
+    function handleMouseClick(e) {
+        // if (!avatarActive) return;
+        spatialIndicatorActivated = true;
+        worldIntersectPoint = getRaycastCoordinates(e.clientX, e.clientY);
+        if (worldIntersectPoint !== undefined) addSpatialIndicator();
+    }
+
+    const indicatorAxis = new THREE.Vector3(0, 1, 0);
+    const indicatorHeight = 400;
+    const indicatorName = 'cylinderIndicator'
```
hobinjk-ptc, line 199: nit: missing semicolon

src/gui/spatialArrow.js
Steve-KX-RL: Good tip. Will keep them within local scope. Thx James!

src/gui/spatialIndicator.js
```diff
@@ -0,0 +1,370 @@
...
+
+        await getMyAvatarColor();
+        uniforms['avatarColor'].value = finalColor;
+    }
+    
+    function updateUniforms() {
+        if (!spatialIndicatorActivated) return;
+        
+        // change the Lines data here
+        lines.forEach(Line => {
```
hobinjk-ptc, line 316: variables should be kept lowercase/switchCase to differentiate from a class

src/gui/spatialArrow.js
hobinjk-ptc: You're welcome! Great work throughout this PR!

src/gui/spatialIndicator.js
```diff
@@ -0,0 +1,370 @@
...
+    const iScaleFactor = 1.1;
+    let indicatorList = [];
+    
+    let innerWidth = 20;
+    let innerBottomHeight = 70;
+    let innerTopHeight = 250;
+    let innerHeightOffset = 50;
+    
+    function addSpatialIndicator() {
+        console.info('should add a cylinder to the scene');
```
dangond-ptc, line 212: We should get rid of these logging statements before committing to avoid clogging up the console.

src/gui/threejsScene.js
```diff
@@ -821,6 +839,38 @@ import { ViewFrustum, frustumVertexShader, frustumFragmentShader, MAX_VIEW_FRUST
         };
     };
 
+    // source: https://stackoverflow.com/questions/29758233/three-js-check-if-object-is-still-in-view-of-the-camera
+    exports.isObjectOnScreen = function(objectPosition) {
```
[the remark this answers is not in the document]
dangond-ptc: It is probably worth noting that this checks if the object's origin is on-screen rather than the entire object. Might be better to rename it `isPointOnScreen`?

src/gui/threejsScene.js
```diff
@@ -848,6 +866,38 @@ import { ViewFrustum, frustumVertexShader, frustumFragmentShader, MAX_VIEW_FRUST
...
+    exports.getToolPosition = function(toolId) {
+        let toolSceneNode = realityEditor.sceneGraph.getSceneNodeById(toolId);
+        let groundPlaneNode = realityEditor.sceneGraph.getGroundPlaneNode();
+        // console.log('%c debugging tool position', 'color: orange');
+        // console.log(realityEditor.sceneGraph.convertToNewCoordSystem({x: 0, y: 0, z: 0}, toolSceneNode, groundPlaneNode));
+        return realityEditor.sceneGraph.convertToNewCoordSystem({x: 0, y: 0, z: 0}, toolSceneNode, groundPlaneNode);
+    }
+
+    // gets the direction the tool is facing, within the coordinate system of the groundplane
+    exports.getToolDirection = function(toolId) {
```
dangond-ptc, line 892: This function and the one above are very useful, thanks for adding them!

src/gui/spatialIndicator.js
Steve-KX-RL: Right! Will delete these logs. Thx Dan!

src/gui/threejsScene.js
Steve-KX-RL: Ben originally added them in the spatial cursor actually. I just moved them into threejsScene.js.

dangond-ptc (approved):

src/gui/spatialArrow.js
```diff
@@ -0,0 +1,224 @@
...
+                drawArrow(finalPosX, finalPosY, angle, 5, indicator.avatarColor, indicator.avatarColorLighter);
+            }
+        })
+    }
+    
+    function drawIndicatorArrows() {
+        searchForIndicators();
+        determineObjectScreenPosition();
+        // draw a single test arrow at the screen center
+
```
benptc, line 215: It looks like the function `determineObjectScreenPosition` is also performing the draw - maybe it can be named more clearly like `drawArrowAtScreenPositionOfEachIndicator` (or something a bit shorter that still gets the point across)

Also, is the comment `// draw a single test arrow at the screen center` referring to the line above it, or is it referring to code that was removed? (in which case it should be removed)

dangond-ptc (approved):

benptc (approved):
Looks good overall to me now that the button-clicking bug is fixed! I left one small comment but that's it for me for now.

hobinjk-ptc (approved):

PR #333, opened 2023-01-26
Never call renderFrame directly; Use requestAnimationFrameIfNotPending

hobinjk-ptc:
Introduces new api to prevent calls to
requestAnimationFrame(renderFrame) before they happen. This prevents an issue where a queue of requestAnimationFrame(renderFrame) would build up from the watchdog during the screen sleeping or other scenarios where setInterval would continue while requestAnimationFrame would pause. While renderFrame has logic to prevent this in theory, in practice the entirety of renderFrame would finish (microtasks) before the next request (macrotask) was popped off the requestAnimationFrame queue.

### DIFF
--- a/src/gui/glRenderer.js
+++ b/src/gui/glRenderer.js
@@ -302,7 +302,9 @@ createNameSpace("realityEditor.gui.glRenderer");
             }
         }
 
-        setTimeout(renderFrame, 500);
+        setTimeout(() => {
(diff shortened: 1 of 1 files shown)


ptc-rdeleeuw:
Looks good to me

ptc-rdeleeuw (approved):

PR #334, opened 2023-01-27
adds ability to setCustomTrashZone in an addon

benptc:
This is necessary to move the trash to the bottom of the screen on the portrait iPhone app, without breaking functionality of the classic app.


hobinjk-ptc:
forget about the classic app, where's the wrath of the lich king app?

dangond-ptc (approved):
Looks good to me!

PR #336, opened 2023-01-30
Moved conversion of conversion of human pose from camera CS to world …

ptc-mklaudiny:
JS receives now human joints in WCS and mm units at the start. All conversions are moved to Swift.

Related to PR in ToolboxApp (https://github.com/ptcrealitylab/vuforia-spatial-toolbox-ios-swift/pull/258).

### DIFF
--- a/src/app/callbacks.js
+++ b/src/app/callbacks.js
@@ -239,67 +239,8 @@ createNameSpace('realityEditor.app.callbacks');
     function receivePoses(pose, timestamp, imageSize) {
 
         let poseInWorld = [];
-        /*
-        let worldObject = realityEditor.worldObjects.getBestWorldObject();
-        if (!worldObject) {
-            console.warn('Could not find world object.');
-            return;
-        }
-        let worldObjectId = worldObject.objectId;
-        let worldNode = realityEditor.sceneGraph.getSceneNodeById(worldObjectId);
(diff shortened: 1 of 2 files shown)


src/app/callbacks.js
```diff
@@ -232,13 +232,14 @@ createNameSpace('realityEditor.app.callbacks');
...
      * Callback for realityEditor.app.getPosesStream
-     * @param {Array<Object>} pose
+     * @param {Array<Object>} pose (in world CS, in mm units)
      * @param {number} timestamp of the pose (in miliseconds, but floating point number with nanosecond precision)
      * @param {Array<number>} [width, height] of the image which the pose was computed from
      */
     function receivePoses(pose, timestamp, imageSize) {
 
         let poseInWorld = [];
+        /*
```
ptc-mklaudiny, line 242: It does not seem that this commented code is necessary after the changes in the app (in particular TODO remarks below). But I want to double-check.

hobinjk-ptc: If you want to make the deletion a separate commit we can refer to the diff of that if anything goes wrong in the future

hobinjk-ptc (approved):
Code LGTM, see ios-swift repo's pr for review based on integrating everything

PR #337, opened 2023-01-30
turn on spatial cursor after localization; dont occlude spatial cursor

benptc:
- Prevents cursor from floating at origin when app is loading / localizing, by waiting until you've localized in a world object and area mesh has downloaded before enabling the cursor.
- Turns off depthTest and depthWrite for the spatial cursor so that it isn't occluded – makes sense for it to show up on top of the area target mesh (especially the invisible occlusion mesh in AR). Prevents z-fighting / other visual artifacts. Just setting the renderOrder doesn't seem to have any effect, so I went with depthTest/depthWrite. Does anyone have a better idea on how to fix this?

### DIFF
--- a/src/spatialCursor/index.js
+++ b/src/spatialCursor/index.js
@@ -4,10 +4,10 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
 
 (function(exports) {
 
-    const SNAP_CURSOR_TO_TOOLS = true;
-    const DEFAULT_SPATIAL_CURSOR_ON = true;
+    const SNAP_CURSOR_TO_TOOLS = false; // the snapping doesn't do anything yet, so turn off for now
+    const DEFAULT_SPATIAL_CURSOR_ON = true; // applied after we localize within a target / download occlusion mesh
 
-    let isCursorEnabled = DEFAULT_SPATIAL_CURSOR_ON;
+    let isCursorEnabled = false; // always starts off, turns on after localize
     let isUpdateLoopRunning = false;
     let occlusionDownloadInterval = null;
     let cachedOcclusionObject = null;
@@ -76,6 +76,7 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         gl_FragColor = vec4(r, g, b, 1.0);
     }
     `;
+    // remember to set depthTest=false and depthWrite=false after creating the material, to prevent visual glitches
     const normalCursorMaterial = new THREE.ShaderMaterial({
         vertexShader: vertexShader,
         fragmentShader: normalFragmentShader,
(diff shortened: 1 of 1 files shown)


hobinjk-ptc:
I think it was being drawn behind the occlusion object which is sort of working as intended. The layers modification is only on desktop. I think this is an okay solution since the cursor isn't expected to be meaningfully behind anything

hobinjk-ptc (approved):

src/spatialCursor/index.js
```diff
@@ -4,10 +4,10 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
 
 (function(exports) {
 
-    const SNAP_CURSOR_TO_TOOLS = true;
-    const DEFAULT_SPATIAL_CURSOR_ON = true;
+    const SNAP_CURSOR_TO_TOOLS = false; // the snapping doesn't do anything yet, so turn off for now
+    const DEFAULT_SPATIAL_CURSOR_ON = true; // applied after we localize within a target / download occlusion mesh
 
-    let isCursorEnabled = DEFAULT_SPATIAL_CURSOR_ON;
+    let isCursorEnabled = false; // always starts off, turns on after localize
```
Steve-KX-RL, line 10: This flag is not used anywhere else. Did you mean to set it to true in the onLoadOcclusionObject callback?

src/spatialCursor/index.js
```diff
@@ -178,18 +177,19 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
...
+
+            if (DEFAULT_SPATIAL_CURSOR_ON) {
+                toggleDisplaySpatialCursor(true);
+            }
         });
         
         addSpatialCursor();
         addTestSpatialCursor();
-        toggleDisplaySpatialCursor(DEFAULT_SPATIAL_CURSOR_ON);
+        toggleDisplaySpatialCursor(false);
```
Steve-KX-RL, line 188: Is this intentional after toggleDisplaySpatialCursor(true) in onLoadOcclusionObject callback? I got a little confused. Sorry!

src/spatialCursor/index.js
benptc: Good thought, but it gets set true in the `toggleDisplaySpatialCursor` function which onLoadOcclusionObject's callback calls. (You might need to unfold some hidden parts of the code in the code review page to see the full file)

Steve-KX-RL:
Agreed with James -- turning off depthTest and depthWrite is a valid solution, since layering is only valid on desktop. We want spatial cursor to be in front of everything, regardless of mobile / desktop.

src/spatialCursor/index.js
benptc: Since the `onLoadOcclusionObject` includes a 1 second interval before it checks whether the occlusion object is loaded, it seems safe to me to add this after the callback. But it could be more robust if we moved these before it.

src/spatialCursor/index.js
Steve-KX-RL: Got it!

PR #338, opened 2023-01-31
Even more analytics api work; Invisible spaghetti finally unhoverable

hobinjk-ptc:
It is likely that regioncards aren't going to exist in a pinned form so some of these changes are going into the wind


dangond-ptc (approved):
Looks good to me!

PR #347, opened 2023-02-10
Refactor createFrame; addToolAtScreenCenter works properly

benptc:
`spatialCursor.addToolAtScreenCenter` adds the tool projected onto the surface of the area target, but at the center of the screen, unless the option `{ moveToCursor: true }` is provided, in which case it will position it at the cursor's position. If the center of the screen doesn't raycast against the area target mesh, it will add it floating in the middle of the screen in front of the camera.

Also cleans up/refactors the pocket.createFrame function to use an options object instead of a million optional parameters, and to remove some deprecated functionality.

### DIFF
--- a/src/envelopeManager.js
+++ b/src/envelopeManager.js
@@ -460,7 +460,14 @@ createNameSpace("realityEditor.envelopeManager");
                 };
 
                 if (envelopeData) {
-                    let addedElement = realityEditor.gui.pocket.createFrame(envelopeData.name, JSON.stringify(envelopeData.startPositionOffset), JSON.stringify(envelopeData.width), JSON.stringify(envelopeData.height), JSON.stringify(envelopeData.nodes), touchPosition.x, touchPosition.y, true);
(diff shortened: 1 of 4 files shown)


src/envelopeManager.js
```diff
@@ -460,7 +460,14 @@ createNameSpace("realityEditor.envelopeManager");
                 };
 
                 if (envelopeData) {
-                    let addedElement = realityEditor.gui.pocket.createFrame(envelopeData.name, JSON.stringify(envelopeData.startPositionOffset), JSON.stringify(envelopeData.width), JSON.stringify(envelopeData.height), JSON.stringify(envelopeData.nodes), touchPosition.x, touchPosition.y, true);
+                    let addedElement = realityEditor.gui.pocket.createFrame(envelopeData.name, {
```
hobinjk-ptc, line 463: Excellent refactor! So much clearer

hobinjk-ptc (approved):

PR #351, opened 2023-02-17
prevent excessive get requests to server from too many heartbeats que…

benptc:
…uing up while the app is in its menus

### DIFF
--- a/src/network/discovery.js
+++ b/src/network/discovery.js
@@ -36,9 +36,9 @@ createNameSpace("realityEditor.network.discovery");
 
     function processNextQueuedHeartbeat() {
         if (queuedHeartbeats.length === 0) { return; }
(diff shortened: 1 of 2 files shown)


hobinjk-ptc (approved):
LGTM! Nice to have more of this setTimeout behavior mitigated

benptc:
> LGTM! Nice to have more of this setTimeout behavior mitigated

Yes! Just realized the existing method was having a huge impact in my new changes to the pop-up app, for various reasons. Particularly, if your phone has multiple world objects saved on its local server, and one of them is designated as the primaryWorld, there would be an endless loop (spaced out my 10ms timeouts) that would make a GET request to the server each time to check whether the other world objects needs to be added. I've circumvented that behavior here.

PR #352, opened 2023-02-23
Humanpose improvements and refactors

hobinjk-ptc:
- InstancedMesh-based rendering
- So many bug fixes
- Dynamic creation of spatialAnalytics tools for each regioncard (will be removed in favor of another approach soon:tm:)
- Misc. functionality improvements like scrolling when at edge of timeline


src/analytics/regionCard.js
```diff
@@ -114,6 +114,19 @@ export class RegionCard {
...
+        const frameKey = addedTool.uuid;
+        const publicData = {
+            startTime: this.startTime,
+            endTime: this.endTime,
+            summary: this.element.outerHTML,
+        };
+        const write = () => {
+            realityEditor.network.realtime.writePublicData(addedTool.objectId, frameKey, frameKey + 'storage', 'status', publicData);
+        };
+        setTimeout(write, 1000);
```
dangond-ptc, line 127: Why does it write twice and with a delay?

dangond-ptc:
Merging our changes is going to be so much fun...

src/analytics/regionCard.js
hobinjk-ptc: The tool added earlier in the function had to load before it starts listening to public data changes so this is a bit paranoid about it

dangond-ptc (approved):
LGTM!

PR #362, opened 2023-03-02
Performance improvements for history spaghetti generation

dangond-ptc:
Now creates spaghetti all at once rather than regenerating it repeatedly one additional point at a time


hobinjk-ptc (approved):
Excellent! Tested locally and LGTM

PR #363, opened 2023-03-02
Terrible hack: disable analytics settings ui on mobile

hobinjk-ptc:


dangond-ptc:
I think you missed two spots (setHistoricalHistoryLinesVisible and the Live version)

hobinjk-ptc:
Yeah I'll close this since I think your proper fix is better

PR #366, opened 2023-03-03
Hack: disable analytics settings ui on mobile

hobinjk-ptc:


benptc (approved):
Seems like a good idea

PR #371, opened 2023-03-06
default floor to bottom of mesh if no geometry directly below; slow down batched updates; safeguard cursor matrix

benptc:
- floorOffset goes to geometry minY if no raycast result (if you didn't scan floor below initial position)
- batchedUpdates happen at only 3fps, to try to minimize network congestion from many avatars, worlds, etc
- avatar keepAlive heartbeat also occurs once per 3 seconds instead of once per 1 second, to reduce residual effects of processing tons of heartbeats. I increased their lifespan to 15 seconds on the server so this balances it here.
- I need to debug the spatial cursor further, but there's a case in which the three.js mesh's matrixWorld is not a properly formatted matrix (its scale values [0, 5, and 10th entry] are all zeros) so the resulting tool placement matrix has a bunch of NaNs in it. I added a safeguard for now to detect this and place the tool in front of the camera rather than being broken until refreshed.

### DIFF
--- a/src/app/navmeshWorker.js
+++ b/src/app/navmeshWorker.js
@@ -268,7 +268,7 @@ const createNavmesh = (geometry, resolution) => { // resolution = number of pixe
     addTriangle(faceData, vertexVector1, vertexVector2, vertexVector3, weight, ignoreWeight);
   }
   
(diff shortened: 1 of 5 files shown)


dangond-ptc (approved):
Looks good to me!

PR #372, opened 2023-03-07
force ignore potential cacheing of the reloadObject and reloadFrame responses

benptc:
On toolboxedge.net, the responses had outdated values, causing frames not to appear. Question: should we default all getData requests to ignore caching, or only turn it on if the option is specified (which I did in only these two usages, to limit the impact of this change)

---
also reverts the batched update framerate back to 10, in case it would have messed up the poses

### DIFF
--- a/src/network/index.js
+++ b/src/network/index.js
@@ -953,7 +953,7 @@ realityEditor.network.onAction = function (action) {
 
                 _this.cout("got object");
 
-            });
+            }, { dontCache: true });
         }
     }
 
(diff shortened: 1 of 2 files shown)


hobinjk-ptc (approved):
LGTM. In general the more caching, the better but it's definitely something we should look into on a case-by-case basis

dangond-ptc (approved):
Looks good to me. I feel that in general, API requests for frequently changing data like we have in our code should avoid caches, but it would likely be easier to make such a change on the cloudflare side than in the code

PR #373, opened 2023-03-07
creates your avatar again if the server deletes it while you were inactive

benptc:
This seems to work in the couple times I've tested it, but ideally I should test it more.

If the server deletes your avatar (tried testing by triggering a breakpoint in my code for 15 seconds), this will cause the client to ask the server to create its avatar again, rather than only creating it one time when the app launches.


hobinjk-ptc (approved):
LGTM

PR #374, opened 2023-03-07
Infinite Grid

dangond-ptc:
These commits should belong to Steve but they needed to be extracted from some unfinished work he had. Adds in the Holodeck-style infinite grid

### DIFF
--- a/src/gui/ar/groundPlaneRenderer.js
+++ b/src/gui/ar/groundPlaneRenderer.js
@@ -88,10 +88,10 @@ createNameSpace("realityEditor.gui.ar.groundPlaneRenderer");
 
         // create an infinite grid that fades into the distance, along the groundplane
         if (!gridHelper) {
-            const colorGrid = new THREE.Color(realityEditor.device.environment.variables.groundWireframeColor);
-            gridHelper = realityEditor.gui.threejsScene.createInfiniteGridHelper(gridSquareSizeInMm, gridRegionSizeInMm, colorGrid, maxVisibilityDistanceInMm);
+            const colorGrid = new THREE.Color('rgb(255, 240, 0)');
+            gridHelper = realityEditor.gui.threejsScene.createInfiniteGridHelper(gridSquareSizeInMm, gridRegionSizeInMm, 0.075, colorGrid, maxVisibilityDistanceInMm);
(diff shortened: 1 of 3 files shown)


src/gui/ar/groundPlaneRenderer.js
```diff
@@ -88,10 +88,10 @@ createNameSpace("realityEditor.gui.ar.groundPlaneRenderer");
 
         // create an infinite grid that fades into the distance, along the groundplane
         if (!gridHelper) {
-            const colorGrid = new THREE.Color(realityEditor.device.environment.variables.groundWireframeColor);
-            gridHelper = realityEditor.gui.threejsScene.createInfiniteGridHelper(gridSquareSizeInMm, gridRegionSizeInMm, colorGrid, maxVisibilityDistanceInMm);
+            const colorGrid = new THREE.Color('rgb(255, 240, 0)');
```
benptc, line 91: Ideally we could still use realityEditor.device.environment.variables.groundWireframeColor, and then just set groundWireframeColor to 'rgb(255, 240, 0)' in the remote operator addon, so that if we ever want to render the groundplane in AR again it will keep its default cyan blue color, but we could go with this for now if we want to quickly merge it in for the upcoming recordings

benptc (approved):
Thanks @Steve-KX-RL for the new visuals!

benptc:
@hobinjk-ptc I'll leave this with you to decide if you want to merge this in and deploy to the cloud version for your upcoming visit, or if you prefer to keep the visuals as-is. This updates the background to look like:

Screen Shot 2023-03-07 at 5 49 33 PM

Steve-KX-RL:
> Thanks @Steve-KX-RL for the new visuals!

Appreciate @dangond-ptc for helping me with this pull request!

src/gui/ar/groundPlaneRenderer.js
Steve-KX-RL: I tried this way, but it seems like it realityEditor.device.environment.variables.groundWireframeColor hasn't been initialized by the time we call it here. But we can fix it later.

Steve-KX-RL (approved):

hobinjk-ptc (approved):
LGTM, it's easy enough to revert if necessary

PR #381, opened 2023-03-16
Remove uv2 attribute from geometries when present

hobinjk-ptc:


benptc (approved):
Thanks for solving this one!

PR #383, opened 2023-03-17
Analytics polish

hobinjk-ptc:
Not tooo happy with the scrollbar on a transparent container, feels like an easy future win to make that look less bad


dangond-ptc (approved):
Looks good to me!

PR #388, opened 2023-03-22
Support resizable window; respond to new onWindowResized tool API

benptc:
Adjust canvas sizes and fullscreen div sizes when the window resizes, and support API that three.js tools can use to listen to changing window size

### DIFF
--- a/src/gui/glRenderer.js
+++ b/src/gui/glRenderer.js
@@ -278,6 +278,13 @@ createNameSpace("realityEditor.gui.glRenderer");
         canvas.style.height = canvas.height + 'px';
         gl = canvas.getContext('webgl2');
 
+        realityEditor.device.layout.onWindowResized(({width, height}) => {
+            canvas.width = width;
+            canvas.height = height;
+            canvas.style.width = width + 'px';
+            canvas.style.height = height + 'px';
+        });
+
         // If we don't have a GL context, give up now
 
         if (!gl) {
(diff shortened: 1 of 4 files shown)


src/gui/glRenderer.js
```diff
@@ -278,6 +278,13 @@ createNameSpace("realityEditor.gui.glRenderer");
         canvas.style.height = canvas.height + 'px';
         gl = canvas.getContext('webgl2');
 
+        realityEditor.device.layout.onWindowResized(({width, height}) => {
+            canvas.style.width = width + 'px';
+            canvas.style.height = height + 'px';
+            // you would think you should update canvas.width and canvas.height too, but that breaks
+            // the coordinate system, for a reason I don't understand. Just update style.width and style.height
```
hobinjk-ptc, line 285: This mostly makes sense given that there's no mechanism for sending the new canvas pixel dimensions (i.e. canvas.width and canvas.height) to the proxied frame. If there were, then the proxied frame could call threejs's setSize function or otherwise update their own viewport to match up. The approach you're taking is a good compromise since most people won't be doing things like resizing from super thin to super wide

hobinjk-ptc (changes requested):
src/analyticas/timeline.js has some calculations based on width that require an onWindowResize callback. Specifically the function draw() has a call to getBoundingClientRect() if (this.width < 0) that will change based on window dimensions. Refactoring that code out into a function that can also be called onWindowResize would fix that handling

hobinjk-ptc:
(one additional note is that there's no need to adjust widthMs based on changes to width in order to keep the same pixels per ms value, this would be a bunch of convoluted logic to get right when just modifying width will get the job done)

src/gui/glRenderer.js
benptc: Thanks for the explanation!

PR #390, opened 2023-03-22
Allow adding region cards during recording

hobinjk-ptc:
Additionally modifies getPosesInTimeInterval to not return duplicate poses where possible. Previously, pulling from clones.all would result in duplicate poses with slightly different timestamps between clones.live and clones.historical.

This new functionality allows the Mark Step button in the spatialAnalytics envelope to function.

### DIFF
--- a/src/humanPose/draw.js
+++ b/src/humanPose/draw.js
@@ -658,7 +658,7 @@ export class HumanPoseAnalyzer {
      * @return {Pose[]} - all poses in the time interval
      */
     getPosesInTimeInterval(firstTimestamp, secondTimestamp) {
-        let clonesList = this.clones.live;
+        let clonesList = this.clones.all;
         if (this.clones.historical.length >= this.clones.live.length) {
             clonesList = this.clones.historical;
         }
--- a/src/humanPose/draw.js
+++ b/src/humanPose/draw.js
@@ -651,13 +651,21 @@ export class HumanPoseAnalyzer {
     }
 
     /**
-     * Sets all poses in the time interval
+     * Returns a list of poses in the time interval, preferring the historical
+     * data source where available
      * @param {number} firstTimestamp - start of time interval in ms
      * @param {number} secondTimestamp - end of time interval in ms
      * @return {Pose[]} - all poses in the time interval
      */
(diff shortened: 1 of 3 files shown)


src/humanPose/draw.js
```diff
@@ -651,13 +651,21 @@ export class HumanPoseAnalyzer {
...
+     * Returns a list of poses in the time interval, preferring the historical
+     * data source where available
      * @param {number} firstTimestamp - start of time interval in ms
      * @param {number} secondTimestamp - end of time interval in ms
      * @return {Pose[]} - all poses in the time interval
      */
     getPosesInTimeInterval(firstTimestamp, secondTimestamp) {
-        const poses = this.clones.all.map(clone => clone.pose).filter(pose => pose.timestamp >= firstTimestamp && pose.timestamp <= secondTimestamp);
+        let clonesList = this.clones.all;
+        if (this.clones.historical.length >= this.clones.live.length) {
```
dangond-ptc, line 662: A bit confused by this logic, why does the relative clone count matter?

dangond-ptc: I think it would make more sense if we had a way to call this function and specify which list to pull from (or two separate functions). It might cause problems if we ever want to get the live poses but there's existing historical data.

hobinjk-ptc: Currently, the one purpose this function serves is to feed into a RegionCard which wants the poses on the timeline between the two times. During recording, this is clones.live. After recording but before the live data is cleared (i.e. page reloaded), this is either clones.live or clones.historical, but not both. After the page is reloaded, this is only clones.historical.

I guess my preferred solution is that clones.historical always contains any clones represented on the timeline and I never have to interact with clones.live.

dangond-ptc: Ohhh, I see. I was wondering when clones.live came into the mix and I forgot you have to get the poses from there in the first place during recording. It makes sense to me now, thanks for taking the time to clear that up for me.

hobinjk-ptc: No problem! The act of explaining this has made me want a better solution so I'm going to take some time to do that

PR #391, opened 2023-03-24
Human tracking multiple cameras

ptc-mklaudiny:
Mostly changes related to live visualisation of multiple human poses which is needed for human tracking in multiple cameras (toolboxapps).    

Related to PR in server repo: https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/701


dangond-ptc (approved):
Looks good to me!

src/humanPose/draw.js
```diff
@@ -1158,25 +1218,29 @@ function updateJointsAndBones(poseRenderInstance, poseObject, timestamp) {
...
+        jointConfidences[jointId] = confidence;
+
+        if (RENDER_CONFIDENCE_COLOR) {
+            
             poseRenderInstance.setJointConfidenceColor(jointId, confidence);
         }
     }
 
-    const pose = new Pose(jointPositions, timestamp, {poseObjectId: poseObject.uuid});
+    const poseHasParent = Boolean(poseObject.parent);
```
hobinjk-ptc, line 1238: It's less pretty but `!!poseObject.parent` is the more standard way of converting to a boolean since `Boolean` is a boxed type

src/humanPose/HumanPoseRenderInstance.js
```diff
@@ -195,9 +195,12 @@ export class HumanPoseRenderInstance {
     }
 
     setVisible(visible) {
+        // MK HACK: too strict to do nothing if the visibility did not change. Other code can 'unhide' the slot
+        /*
```
hobinjk-ptc, line 199: I believe this is a pretty sizeable performance loss but I'll do some proper testing to confirm. I definitely think this optimization is here for a reason

ptc-mklaudiny: The problem was that unwanted skeletons appeared in some cases. The slot of this render instance got changed (showSlot/hideSlot) but the visibility of this instance was not updated accordingly. Then the early exit did not give it a chance to correct itself.

But we spoke with @dangond-ptc and thought that one could just delete the instance object entirely instead for hiding it (by collapsing whole skeleton to 0,0,0).  There is a comment in function `hidePoseRenderInstance(poseRenderInstance) {...}`

hobinjk-ptc: Yes, there's poseRenderInstance.remove() which will free the slot on the renderer. hidePoseRenderInstance could call remove() and set poseRenderInstances[poseRenderInstance.id] to null. I don't think this would fix whatever underlying behavior is causing this mismatch between poseRenderInstance.visible and the pose render instance's actual visibility.

hobinjk-ptc: As promised, I did profile the impact of this change and it appears to be minimal (compared to all the other parts that need to be optimized). It would be better to fix the underlying issue but I'm okay with this hack for now

ptc-mklaudiny: Thanks.

ptc-mklaudiny:
I made another round of changes to allow toggle visualisation of auxiliary poses (from standard human objects which support fused human objects at a given timestamp). This is not supported in live, live history and history playback visualisation. To get it working properly, I had to remove some early exits (see MK HACK comments). 
I could not find underlying cause why these optimisations do not work (eg. in live mode when one is toggling 'View auxiliary poses' in analytics settings). Someone more knowledgeable of this code needs to have a look how to bring those optimisations back (this needs at least two phones).
However, I would like to merge this and related server PR soon so the team can test more multi-camera human tracking. So these further fixes could be left for later.

hobinjk-ptc (approved):
Good to have this! Definitely excited to work on this improved foundation

src/humanPose/draw.js
```diff
@@ -273,9 +275,12 @@ export class HumanPoseAnalyzer {
      * @param {Pose} pose - the pose renderer that was updated
      * @param {boolean} historical - whether the pose is historical or live
      */
-    poseUpdated(pose, historical) {
+    poseUpdated(pose, historical, skipLine) {
```
hobinjk-ptc, line 278: Unused parameter to remove here

dangond-ptc (approved):
Looks good to me! Agree with James, it's exciting to have better support for multiple cameras now!

PR #397, opened 2023-03-30
Change tool open close animations

Steve-KX-RL:


src/gui/ar/draw.js
```diff
@@ -2245,33 +2245,26 @@ realityEditor.gui.ar.draw.removeFullscreenFromFrame = function(objectKey, frameK
...
-            // also fade it out and remove it entirely when done
-            envelopeAnimationDiv.style.opacity = 0;
-            setTimeout(function() {
-                envelopeAnimationDiv.parentElement.removeChild(envelopeAnimationDiv);
-            }, 250);
+
+        const parentDiv = globalDOMCache['object' + frame.uuid];
+        let tempAnimDiv = document.createElement('div');
+        tempAnimDiv.classList.add('temp-anim-div');
+        tempAnimDiv.style.transform = 'matrix3d(643.374, -0.373505, 0.000212662, 0.000212647, 0.372554, 643.38, 0.000554764, 0.000554727, -2.77404, 4.28636, 0.500033, 0.5, -1406.67, 2173.54, 34481.6, 253.541)';
```
benptc, line 2252: Can you add a comment explaining how you got this hard-coded matrix, in case someone else wants to modify it in the future?

benptc (approved):
I tested it out and it looks good! Can you just add a comment or two explaining how it works? After that we can merge the PR

src/gui/ar/draw.js
Steve-KX-RL: I added a tool, closed it to reveal the icon, and moved the camera towards the tool, so that the tool almost fills up the screen in the center. And then I get the matrix3d of the object that the tool is attached to. Very hacky, hope to make it procedural in the future😆.

PR #398, opened 2023-04-03
Reliably sync envelope state

benptc:
Goes with https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/748

Fixes for envelope reliability:
- don't redundantly open/close envelope if already in open or close state
- refactor initNode and createNode tool API message handlers into a single implementation, which also accepts a `defaultValue` for the node.

Other bug tracked down:
- Minimize initialization messages sent into iframes by not calling `realityEditor.network.onElementLoad` on ALL tools when receiving an updateObject action message, only on tools using the screen<>AR visualization experiment (as that was the feature that this code was added).

### DIFF
--- a/src/envelopeManager.js
+++ b/src/envelopeManager.js
@@ -9,11 +9,6 @@ createNameSpace("realityEditor.envelopeManager");
  */
 
 (function(exports) {
-    
-    const DEBUG_LOG = true;
-    function log(str) {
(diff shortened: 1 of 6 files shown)


dangond-ptc (approved):
Looks good to me!

hobinjk-ptc (approved):
Nice reorganization!

PR #401, opened 2023-04-05
poseHasParent should be false when parent is undefined or null

hobinjk-ptc:
Single camera and old historical pose data would have no parent which `!= "none"` causing the new logic to erroneously set poseHasParent to true.

PR also removes trailing whitespace


ptc-mklaudiny (approved):
Oh, right. That broke backwards compatibility with recorded Mercedes data.

PR #403, opened 2023-04-07
Updated AccelerationLens to work with multiple poses and use m/s^2 co…

dangond-ptc:
…rrectly

### DIFF
--- a/src/humanPose/AccelerationLens.js
+++ b/src/humanPose/AccelerationLens.js
@@ -58,8 +58,7 @@ export class AccelerationLens extends AnalyticsLens {
                 const previousJoint = previousPose.getJoint(joint.name);
                 joint.velocity = joint.position.clone().sub(previousJoint.position).divideScalar(pose.timestamp - previousPose.timestamp); // mm/ms = m/s
                 joint.speed = joint.velocity.length();
-                const previousPreviousJoint = previousPreviousPose.getJoint(joint.name);
-                joint.acceleration = joint.velocity.clone().sub(previousPreviousJoint.velocity).divideScalar((pose.timestamp - previousPose.timestamp) / 1000); // Divide by 2 to get the average time between the two poses, and divide by 1000 to convert from ms to s
+                joint.acceleration = joint.velocity.clone().sub(previousJoint.velocity).divideScalar((pose.timestamp - previousPose.timestamp) / 1000);
                 joint.accelerationMagnitude = joint.acceleration.length();
             });
         }
--- a/src/humanPose/AccelerationLens.js
+++ b/src/humanPose/AccelerationLens.js
@@ -3,8 +3,8 @@ import {AnalyticsLens} from "./AnalyticsLens.js";
 import {AnalyticsColors} from "./AnalyticsColors.js";
 import {JOINTS} from "./utils.js";
 
-const HIGH_CUTOFF = 35000 // In ???/???^2
-const MED_CUTOFF = 15000 // In ???/???^2
+const HIGH_CUTOFF = 35 // In m/s^2
+const MED_CUTOFF = 15 // In m/s^2
 
 /**
  * AccelerationLens is a lens that calculates the acceleration of each joint in the pose history.
@@ -15,15 +15,6 @@ export class AccelerationLens extends AnalyticsLens {
      */
     constructor() {
         super("Acceleration");
-        
-        // For live rendering
-        this.previousPose = null;
-        this.previousPreviousPose = null;
-    }
-    
-    reset() {
-        this.previousPose = null;
-        this.previousPreviousPose = null;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/humanPose/draw.js
```diff
@@ -1256,6 +1258,10 @@ function updateJointsAndBones(poseRenderInstance, poseObject, timestamp) {
 
     const poseHasParent = poseObject.parent && (poseObject.parent !== 'none');
     const pose = new Pose(jointPositions, jointConfidences, timestamp, {poseObjectId: poseObject.uuid, poseHasParent: poseHasParent});
+    pose.metadata.previousPose = mostRecentPoseByObjectId[poseObject.uuid];
+    if (!poseHasParent) {
```
ptc-mklaudiny, line 1262: I would not restrict acceleration calculation to just human objects without parent. It can be available to all of them regardless whether they are auxiliary.

src/humanPose/index.js
```diff
@@ -240,6 +241,10 @@ import {JOINT_TO_INDEX} from './constants.js';
                     poseObjectId: identifier,
                     poseHasParent: poseObject.parent && (poseObject.parent !== 'none'),
                 });
+                pose.metadata.previousPose = mostRecentPoseByObjectId[poseObject.uuid];
+                if (!pose.metadata.poseHasParent) {
```
ptc-mklaudiny, line 245: Same as above.

src/humanPose/draw.js
```diff
@@ -1256,6 +1258,10 @@ function updateJointsAndBones(poseRenderInstance, poseObject, timestamp) {
 
     const poseHasParent = poseObject.parent && (poseObject.parent !== 'none');
     const pose = new Pose(jointPositions, jointConfidences, timestamp, {poseObjectId: poseObject.uuid, poseHasParent: poseHasParent});
+    pose.metadata.previousPose = mostRecentPoseByObjectId[poseObject.uuid];
```
ptc-mklaudiny, line 1261: Just a side question: Where HumanPoseObject or generic ObjectModel get uuid property? In server code, they are defined just with objectId.

ptc-mklaudiny (changes requested):

src/humanPose/draw.js
dangond-ptc: Do the auxiliary poses have their own poseObjectIds? If so, how do we properly tell which pose corresponds to the previous pose of each pose? I was under the impression that the auxiliary poses share an object Id, and jumping between them is what we're trying to avoid

ptc-mklaudiny: Yes, they do.  An example from my console: _HUMAN_server_Wajvnee6619_pose1__Igya99zswgi is 'fused' from _HUMAN_devicekz1gkwfb_pose1__peakszmzux1 and _HUMAN_devicej1O8vomf_pose1__Tcmbg2g9kvl (both have parent property set to  _HUMAN_server_Wajvnee6619_pose1__Igya99zswg)

src/humanPose/draw.js
dangond-ptc: I think this is a holdover from old code, I'll replace it with objectId everywhere in my next commit

src/humanPose/AccelerationLens.js
```diff
@@ -45,37 +36,32 @@ export class AccelerationLens extends AnalyticsLens {
...
+                const previousJoint = previousPose.getJoint(joint.name);
+                joint.velocity = joint.position.clone().sub(previousJoint.position).divideScalar(pose.timestamp - previousPose.timestamp); // mm/ms = m/s
                 joint.speed = joint.velocity.length();
-                const previousPreviousJoint = this.previousPreviousPose.getJoint(joint.name);
-                joint.acceleration = joint.velocity.clone().sub(previousPreviousJoint.velocity).divideScalar(((pose.timestamp - this.previousPreviousPose.timestamp) / 2) / 1000); // Divide by 2 to get the average time between the two poses, and divide by 1000 to convert from ms to s
+                const previousPreviousJoint = previousPreviousPose.getJoint(joint.name);
+                joint.acceleration = joint.velocity.clone().sub(previousPreviousJoint.velocity).divideScalar((pose.timestamp - previousPose.timestamp) / 1000); // Divide by 2 to get the average time between the two poses, and divide by 1000 to convert from ms to s
```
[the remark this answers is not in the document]
dangond-ptc: Thanks, when I changed the time interval I forgot to change the source velocity, completely agree with you on that

src/humanPose/draw.js
ptc-mklaudiny: In each timestamp, we reason whether the pose for the fused 'server' object should be taken from one or the other auxiliary object. But they are all fully-fledged human pose object with own pose history.

hobinjk-ptc (approved):
Nice simplifications! Storing previousPose on the metadata is feels solid. Looking forward to having concrete units

ptc-mklaudiny (approved):
Thanks for the improvements.

PR #404, opened 2023-04-07
trigger onLocalizedWithinWorld immediately if already localized

benptc:
Might be responsible for some small bugs, if we've already localized before a module subscribes to the listener


hobinjk-ptc (approved):
A classic improvement to make, LGTM!

PR #406, opened 2023-04-10
Add recently used envelopes bar

hobinjk-ptc:


benptc (approved):
Very slick! LGTM!

benptc:
I didn't test this on iOS yet, and I can imagine we might have some issues with responsive design if the screen is too small, but that's something we can polish later

hobinjk-ptc:
Oh, good point! I'll do a quick test to make sure it isn't unbearable

hobinjk-ptc:
Could be better but I'm declaring it a good point from which to improve

PR #409, opened 2023-04-11
fix issue with onLocalized UI getting stuck localizing forever

benptc:
The previous fix that I was using to get `onLocalizedWithinWorld` to trigger wasn't always working; it seems to be because if you're still on the Setup Metaverse screen when it localizes, the toolbox has a `suppressObjectRendering` flag toggled on, causing `getBestWorldObject` used internally in onLocalizedWithinWorld  to return the _WORLD_local.

This new method has a backup, where it returns the world ID directly used by the `setOrigin` function.

### DIFF
--- a/src/worldObjects.js
+++ b/src/worldObjects.js
@@ -25,6 +25,7 @@ createNameSpace("realityEditor.worldObjects");
      * @type {Object.<string, Array.<number>>}
      */
     var worldCorrections = {};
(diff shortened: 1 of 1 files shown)


hobinjk-ptc (approved):
Nice! Just ran into this bug earlier today

Steve-KX-RL (approved):
Nice! Tested working!

PR #412, opened 2023-04-14
Updated addToolAtScreenCenter() to work with the new off-center camera system.

Steve-KX-RL:
Works in conjunction with the new remote operator rotateCenterElement update.


benptc (approved):
LGTM

PR #413, opened 2023-04-17
X button only closes the focused envelope; increase default tool scale from 0.5 to 1

benptc:
- The [X] button and the minimize [-] button only show if there's an active focused envelope
- Pressing the [X] button only closes the active focused envelope
- Selecting a recently-used tool only closes the focused envelope; also restores focus to the one you select
- Increase the defaultScale of tools from 0.5 to 1, so that the tool icons are more noticeable
- Refactors the updateExitButton function

### DIFF
--- a/src/envelopeManager.js
+++ b/src/envelopeManager.js
@@ -10,9 +10,6 @@ createNameSpace("realityEditor.envelopeManager");
 
 (function(exports) {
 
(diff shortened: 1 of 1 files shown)


hobinjk-ptc (approved):
Nice! LGTM

PR #414, opened 2023-04-17
Support minimizing multiple spatialAnalytics tools

hobinjk-ptc:
This PR supports spatialAnalytics tool minimizing where multiple tools can all load their own history but only the active one controls and interacts with the timeline.

The main change to enable this was making HumanPoseAnalyzer its own file with a bit more compartmentalization. A few functions in src/humanPose/draw.js, most notably the live pose rendering, poke holes in this abstraction but overall it's working in a mostly straightforwards one HPA = one envelope model. I considered extending the current split of live + historical into live + Array<historical> but felt this was a cleaner design overall.

I have not yet tested this with multiple phones

### DIFF
--- a/src/analytics/analytics.js
+++ b/src/analytics/analytics.js
@@ -4,12 +4,14 @@ import {
     RegionCardState,
 } from './regionCard.js';
 import {
+    setActiveFrame,
     setHighlightRegion,
     setDisplayRegion,
     setCursorTime,
     clearHistoricalData,
     showAnalyzerSettingsUI,
-    hideAnalyzerSettingsUI, getPosesInTimeInterval
+    hideAnalyzerSettingsUI,
+    getPosesInTimeInterval,
 } from '../humanPose/draw.js';
 import {
     loadHistory
@@ -20,6 +22,8 @@ import {
 
 export class Analytics {
     constructor() {
+        this.activeEnvelope = null;
+
         this.container = document.createElement('div');
         this.container.id = 'analytics-container';
 
@@ -28,7 +32,7 @@ export class Analytics {
 
         this.container.appendChild(this.timelineContainer);
         this.timeline = new Timeline(this.timelineContainer);
-        this.added = false;
+        this.opened = false;
         this.loadingHistory = false;
         this.livePlayback = false;
         this.pinnedRegionCards = [];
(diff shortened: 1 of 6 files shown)


dangond-ptc (approved):
Looks good to me, but Martin should test this with multiple cameras before we merge to be safe.

hobinjk-ptc:
Requires https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/768 and https://github.com/ptcrealitylab/vuforia-spatial-core-addon/pull/238 for tracking active spatial analytics tool

src/analytics/analytics.js
```diff
@@ -41,25 +45,69 @@ export class Analytics {
...
+     */
+    focus(frameId) {
+        this.open(frameId);
+    }
+
+    /**
+     * On envelope blur
+     * Remove all 2d ui
+     */
+    blur(frameId) {
```
ptc-mklaudiny, line 102: I understand  focus/blur naming but at least blur is confusing when encountered in the code somewhere. Maybe foreground/background is clearer pairs but they are not verbs.

src/humanPose/draw.js
```diff
@@ -1463,14 +439,17 @@ function setHumanPosesVisible(visible) {
  * @param {boolean} visible - whether to show or not
  */
 function setChildHumanPosesVisible(visible) {
```
ptc-mklaudiny, line 441: This feels like it should just affect the active HPA. Also, in UI it is a toggle in its own settings.

src/humanPose/draw.js
```diff
@@ -1454,7 +426,11 @@ function toggleAnalyzerSettingsUI() {
  * @param {boolean} visible - whether to show or not
  */
 function setHumanPosesVisible(visible) {
```
ptc-mklaudiny, line 428: This would be better across all HPAs/analytics tools, it is in the top menu bar. In terms of broad semantics, toggle for 3D video in that menu also switches on/off for all avatars

src/humanPose/HumanPoseAnalyzer.js
```diff
@@ -0,0 +1,1129 @@
...
+    get activeLens() {
+        return this.lenses[this.activeLensIndex];
+    }
+
+    /**
+     * Sets up the containers for the history lines and clones
+     * @param {Object3D} parent - object to add the analyzer's containers to
+     */
+    setupContainers(parent) {
+        this.historicalHistoryLineContainer = new THREE.Group();
```
ptc-mklaudiny, line 137: Whenever I see historicalHistory it makes me smile. How about playback instead? For example, historicalHistoryLineContainer  -> playbackHistoryLineContainer  (Its counterpart being liveHistoryLineContainer)

ptc-mklaudiny (approved):
It makes sense to me in general. Few small suggestions left inline

src/humanPose/HumanPoseAnalyzer.js
dangond-ptc: I've felt the same way and like this suggestion. Makes it much easier to understand

src/analytics/analytics.js
ptc-mklaudiny: I have not realised that it is actually used more broadly. It is probably not worth refactoring before Liveworx.  However, it may be an improvement task in backlog to see if better naming can be found.

src/humanPose/HumanPoseAnalyzer.js
hobinjk-ptc: I agree that playback is a better name but renaming this variable would kick off a cascade of historical -> playback that I leave as an exercise for the future.

dangond-ptc: I can tackle that after Liveworx

PR #416, opened 2023-04-18
Add basic vehicleDeleted handling to RecentlyUsedBar

hobinjk-ptc:


benptc (approved):
Looks good! Made me realize there's a small bug in my last changes to the envelope manager, when I refactored the updateExitButton. I'll make a PR for that soon.

PR #419, opened 2023-04-19
Enable human tracking on opening spatialAnalytics tool; Mobile fixes

hobinjk-ptc:


ptc-mklaudiny (commented):
It seems sensible. What happens when I have already analytics tool open in remote viewer and I connect new toolbox app?
I assume that new app by default does not do pose tracking. Internally, it starts in that state and the analytics tool is opened so no msg comes from there.
I could switch it on the device through the new toggle in settings. Is that code under some other PR?

hobinjk-ptc:
The app will localize to the world object then load the spatialAnalytics tool, automatically opening it since it's open in the remote viewer. This should then turn on human tracking.

There is also https://github.com/ptcrealitylab/pop-up-onboarding-addon/pull/42 for the manual toggle

ptc-mklaudiny (approved):
OK, let's try it out. It should allow for playing with special cases for debugging purposes.

PR #420, opened 2023-04-19
Animated lines from recently used icons to their corresponding tool

benptc:
After a chat with Valentin, I increased the speed of the animation to 60ms (can modify `this.hoverAnimationDurationMs = 60;` to change this), down from the ~160ms I tested with earlier, to make the animation super subtle. We can adjust this later if we want the animation more or less visible.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -2007,6 +2007,14 @@ Logic Interface CSS
     border-color: white;
 }
 
+.ru-canvas {
+    position: absolute;
+    left: 0;
+    top: 0;
+    width: 100vw;
+    height: 100vh;
+}
+
 .iframeOverlayWithoutFocus {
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -2007,6 +2007,14 @@ Logic Interface CSS
     border-color: white;
 }
 
+.ru-canvas {
+    position: absolute;
```
hobinjk-ptc, line 2011: Should have `pointer-events: none` for a bit of paranoia

hobinjk-ptc (approved):
Nice! LGTM

css/index.css
benptc: Great point – too many bugs have been caused in the past by forgetting to include that style

PR #424, opened 2023-04-20
Make minimized icons green; load svgs instead of gifs

benptc:
After adding https://github.com/ptcrealitylab/pop-up-onboarding-addon/pull/43 and https://github.com/ptcrealitylab/vuforia-spatial-core-addon/pull/241, loads the white-line-only icon-foreground.svg files instead of the black-background icon.gifs, and adds a green background with CSS, to make the minimized icons stand out more against the background


hobinjk-ptc (approved):
Code LGTM, I'm less sure about the PTC green though :D

benptc:
Agreed about the green – updated it to have a black background, white outline, and the same color animation as the existing tool icons

PR #426, opened 2023-04-21
Allow clicking on icons while drawing is open

benptc:
Moves fullscreen tools to background (z = -10) so we can click on tool icons. If fullscreen tool is full2D (chat and search), moves to foreground (z = 100) so the icons don't render on top of the 2D content.

Since new envelopes can be added/opened while one is focused now, added some code to the focus handlers to close/blur the active envelope when a new envelope enters the foreground.

Also adds a hover visual feedback to minimized icons, so you can tell when you will click on them vs the background.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -2020,9 +2020,19 @@ Logic Interface CSS
     pointer-events: none; /* deactivates the tool's touchPropagation cover when an envelope loses focus */
 }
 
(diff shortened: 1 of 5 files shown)


hobinjk-ptc (approved):
Nice! Tested locally and LGTM

PR #428, opened 2023-04-21
Fixed iOS add tool when spatial cursor outside of AreaTarget mesh bug

Steve-KX-RL:


benptc (approved):
LGTM! Seems to be working!

PR #430, opened 2023-04-21
if a toolId is present in the URL, attempts to open the envelope when it loads

benptc:
There are some race conditions if you have multiple minimized envelopes and also pass a toolId into the URL. I ran out of brainpower to solve this without adding a timeout before opening, and another timeout before focusing. Probably not 100% robust if on a slow network, but works in the cases where I tried it. In the worst case, it will just open minimized, or close entirely after opening, if another envelope with an open=1 node loads too slowly.


hobinjk-ptc (approved):
A reasonable attempt!

PR #433, opened 2023-04-24
Disable loadHistory on mobile; Respect recordingClones; Reduce max poses on mobile

hobinjk-ptc:
Cherry-picked this out of my ongoing analytics work to provide a fix for poses not appearing on mobile (it should work standalone but I've only tested it with all of my other unrelated work)

### DIFF
--- a/src/humanPose/HumanPoseAnalyzer.js
+++ b/src/humanPose/HumanPoseAnalyzer.js
@@ -280,7 +280,7 @@ export class HumanPoseAnalyzer {
      * @param {boolean} historical - whether the pose is historical or live
      */
     poseUpdated(pose, historical) {
-        if (this.recordingClones) {
+        if (this.recordingClones || historical) {
             this.addCloneFromPose(pose, historical);
         }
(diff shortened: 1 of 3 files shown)


src/humanPose/HumanPoseAnalyzer.js
```diff
@@ -273,7 +280,9 @@ export class HumanPoseAnalyzer {
      * @param {boolean} historical - whether the pose is historical or live
      */
     poseUpdated(pose, historical) {
-        this.addCloneFromPose(pose, historical);
+        if (this.recordingClones) {
+            this.addCloneFromPose(pose, historical);
```
dangond-ptc, line 284: Shouldn't we also add the pose if it's historical, regardless of `recordingClones`? I thought `recordingClones` was meant for live data.

hobinjk-ptc: That makes sense, I'll adjust accordingly

dangond-ptc (approved):

PR #434, opened 2023-04-24
adds StackingOrder module to project, to manage different layers/stackingContexts in the dom

benptc:
I didn't manage to finish this up and get to the point where it actually fixes the issue of the drawing app's z-index on Safari, but I've put into place the beginnings of a system to better manage the stacking order of the application.

The stackingOrder creates a number of different layers in the app (background / scene / GUI, with more granularity), and all HTML elements that were added directly to the body are now assigned (and appended, instead) to one of these layers.

By giving each stacking layer a z-index, it establishes a separate stacking context for each layer, so that an element in a lower layer (theoretically) could never render above an element from a higher layer, and vice-versa.

Some more work may be needed to update the CSS for the various elements that are now part of these layers, or eliminate now-redundant/unnecessary CSS for some elements.

Additionally, I still want to move the tools from the GUI element in the `SCENE_APP_ICONS_AND_WIDGETS` to either the `SCENE_APP_BACKGROUND` or the `GUI_APP_FOREGROUND` depending on if they open in fullscreenFull2D mode or not (chat app would go to foreground layer, drawing app would go to background layer). This should fix the issue in safari where the drawing needs to be sandwiched above the scan but below the icons, while the chat app is above everything.

### DIFF
--- a/src/gui/StackingOrder.js
+++ b/src/gui/StackingOrder.js
@@ -49,7 +49,7 @@ class StackingOrder {
 
         this.addToLayer(document.getElementById('glcanvas'), LAYERS.SCENE_APP_BACKGROUND);
 
-        this.addToLayer(document.querySelector('.canvas-node-connections'), LAYERS.SCENE_FOREGROUND);
+        this.addToLayer(document.querySelector('.canvas-node-connections'), LAYERS.SCENE_APP_BACKGROUND);
         this.addToLayer(document.getElementById('p5WebGL'), LAYERS.SCENE_FOREGROUND);
         this.addToLayer(document.getElementById('groupLassoSVG'), LAYERS.SCENE_FOREGROUND);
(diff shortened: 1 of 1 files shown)


hobinjk-ptc:
I think this is a good start! Taking a look at https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context this feels like a pretty ideal approach

benptc:
I was planning to move the tools' iframes to different parent elements when they open/close to adjust which stacking context they belong to (moving chat to foreground, drawing tool to background, keeping tool icons in the middle). But this has run into an unexpected issue. It seems there is no way to prevent an iframe from reloading when it is appended to a different part of the DOM tree (https://stackoverflow.com/a/39895407). So when I tap on the chat icon, opening it fullscreen, and then move it to the foreground layer, the iframe reloads and it reverts back to its initial icon state.

PR #435, opened 2023-04-25
Disable limitScanRAM functionality

hobinjk-ptc:
With improvements to Vuforia's AT scanning functionality, it is now very difficult for us to run out of RAM during the scanning process.  This removes the limitScanRAM function which, while disabled by default, could be previously toggled on on older devices.

### DIFF
--- a/src/gui/ar/areaTargetScanner.js
+++ b/src/gui/ar/areaTargetScanner.js
@@ -23,7 +23,7 @@ createNameSpace("realityEditor.gui.ar.areaTargetScanner");
 
     let hasFirstSeenInstantWorld = false;
 
-    let limitScanRAM = false; // if true (toggled through menu), stop area target capture when device memory usage is high
+    const limitScanRAM = false; // if true, stop area target capture when device memory usage is high
     let maximumPercentRAM = 0.33; // the app will stop scanning when it reaches this threshold of total device memory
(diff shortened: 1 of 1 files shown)


src/gui/ar/areaTargetScanner.js
```diff
@@ -115,18 +115,19 @@ createNameSpace("realityEditor.gui.ar.areaTargetScanner");
...
-            function(newValue) {
-                console.log('limitScanRAM was set to ' + newValue);
-                limitScanRAM = newValue;
-            },
-            function(newValue) {
-                console.log('zone text was set to ' + newValue);
-                maximumPercentRAM = parseFloat(newValue) || 0.33;
-            }
-        ).moveToDevelopMenu();
+        // realityEditor.app.subscribeToAppMemoryEvents('realityEditor.gui.ar.areaTargetScanner.onAppMemoryEvent');
```
dangond-ptc, line 118: Shouldn't we just delete this rather than comment it out?

dangond-ptc (approved):
LGTM!

src/gui/ar/areaTargetScanner.js
hobinjk-ptc: Good point! Updated accordingly

PR #438, opened 2023-04-28
Create new Analytics for every spatialAnalytics tool

hobinjk-ptc:
Simplifies AnalyticsMobile implementation
Fixes various state issues with focus/blur/open/close
Enables reuse of previously loaded data (no more reloading the entire mercedes dataset every time you open the tool)

### DIFF
--- a/src/analytics/index.js
+++ b/src/analytics/index.js
@@ -77,19 +77,21 @@ import {AnalyticsMobile} from './AnalyticsMobile.js'
         });
 
         realityEditor.network.addPostMessageHandler('analyticsFocus', (msgData) => {
-            if (!analyticsByFrame[msgData.frame] || activeFrame === msgData.frame) {
+            if (!analyticsByFrame[msgData.frame]) {
                 return;
             }
-            getActiveAnalytics().blur();
+            if (activeFrame !== msgData.frame) {
+                getActiveAnalytics().blur();
+            }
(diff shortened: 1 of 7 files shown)


src/analytics/index.js
```diff
@@ -4,74 +4,133 @@ import {Analytics} from './analytics.js'
...
+     */
+    function makeAnalytics(frame) {
+        if (realityEditor.device.environment.isDesktop()) {
+            return new Analytics(frame);
+        } else {
+            return new AnalyticsMobile(frame);
         }
+    }
+
+    const noneFrame = 'none';
```
dangond-ptc, line 19: Is there a reason to use this instead of `null`?

dangond-ptc (approved):
Aside from one small nit, looks good to me! Really glad to see timelines and HPAs more strongly managed by a single frame under one Analytics object, makes it much easier to reason about what's going on.

src/analytics/index.js
hobinjk-ptc: The use of a string here is entirely because it felt weird to index by null

dangond-ptc: Fair enough

hobinjk-ptc: If you prefer, I can make up something about how using null would trigger an engine-level deoptimization of the underlying javascript map

PR #440, opened 2023-05-01
Analytics opacity adjustments

hobinjk-ptc:


src/humanPose/HumanPoseAnalyzer.js
```diff
@@ -14,7 +14,7 @@ import {HumanPoseRenderer} from './HumanPoseRenderer.js';
 import {HumanPoseRenderInstance} from './HumanPoseRenderInstance.js';
 import {MAX_POSE_INSTANCES, MAX_POSE_INSTANCES_MOBILE} from './constants.js';
 
-const POSE_OPACITY_BASE = 0.5;
+const POSE_OPACITY_BASE = 0.9;
```
ptc-mklaudiny, line 17: Why not just 1.0 to prevent any background bleeding in?

ptc-mklaudiny: Further question is why not use opaqueRenderer as for markStart/End?  But I guess POSE_OPACITY_BASE is more widely used now, so it affects several visualisation modes for 'live history'/'historical history' poses.

hobinjk-ptc: Fair enough, will update to 1.0. The way selection is handled is first and last poses are on opaquePoseRenderer which remains opaque and the middle selected poses are on live/historicalPoseRenderer (where the opacity is temporarily lowered to POSE_OPACITY_BACKGROUND)

ptc-mklaudiny (approved):

PR #442, opened 2023-05-01
Analytics spaghetti refactor

dangond-ptc:
Refactors the spaghetti code to handle multiple spaghetti lines under one object which keeps track of selection state. Fixes issues where large selections on the spaghetti would not be recognized. Also changes coloring of spaghetti to fade out points that are not within the selection region. Highlighting points also uses the original point's color now.


src/gui/threejsScene.js
```diff
@@ -549,11 +549,6 @@ import { ViewFrustum, frustumVertexShader, frustumFragmentShader, MAX_VIEW_FRUST
 
         raycaster.firstHitOnly = true; // faster (using three-mesh-bvh)
 
-        // add object layer to raycast layer mask
-        objectsToCheck.forEach(obj => {
-            raycaster.layers.mask = raycaster.layers.mask | obj.layers.mask;
-        });
-
```
benptc, line 556: Is there a reason this is removed? I remember adding it for some reason but now I forget exactly why.

benptc (approved):
Nice work!

src/gui/threejsScene.js
dangond-ptc: It's removed because there is another line right below where the exact same piece of code is run redundantly.

benptc: Good catch! 🕵️

src/gui/ar/meshPath.js
```diff
@@ -351,15 +391,16 @@ export class MeshPath extends THREE.Group
...
-            let bufferIndices = this.getBufferIndices(index);
-            bufferIndices.forEach(bfrIndex => {
+            let colorBufferIndices = this.getBufferIndices(index, COMPONENTS_PER_COLOR);
+            colorBufferIndices.forEach(bfrIndex => {
                 let newColor = {
                     r: this.currentPoints[index].color[0],
                     g: this.currentPoints[index].color[1],
-                    b: this.currentPoints[index].color[2]
+                    b: this.currentPoints[index].color[2],
+                    a: this.currentPoints[index].length === 3 ? 255 : this.currentPoints[index].color[3]
```
hobinjk-ptc, line 400: I believe this should be currentPoints[index].color.length

dangond-ptc: Good catch!

hobinjk-ptc (approved):

PR #443, opened 2023-05-02
Add tool at specified location

ktheo-ptc:
- new functions added to place new tools at specific locations in model

duplicates of existing functions were created to avoid conflicts in other parts of the codebase. addToolAtSpecifedCoords could eventually be merged with addToolAtScreenCenter.
- addToolAtSpecifiedCoords is a copy of addToolAtScreenCenter
- getOrientedCursorAtSpecificCoords is a copy of getOrientedCursorIfItWereAtScreenCenter


benptc (approved):
I think in the long term we probably want to refactor these into a more generic method that can add tools in either the center, the cursor position, or a screen X,Y, so as not to duplicate so much code – but for now I'm ok with this solution, as it is all working

PR #444, opened 2023-05-02
Alter three's updateMatrixWorld to not touch hidden objects

hobinjk-ptc:
Significantly improves the performance of updateMatrixWorld, previously a large contributor to frame time.

### DIFF
--- a/thirdPartyCode/three/three.module.js
+++ b/thirdPartyCode/three/three.module.js
@@ -8105,6 +8105,9 @@ class Object3D extends EventDispatcher {
 	}
 
 	updateMatrixWorld( force ) {
+		if (!this.visible) {
+			return;
+		}
 
 		if ( this.matrixAutoUpdate ) this.updateMatrix();
 


dangond-ptc:
How does this change work if you change the matrices of something that is invisible and then make it visible? Is this ever a problem?

hobinjk-ptc:
As far as I can tell what will happen is the object's matrices will then be calculated on the frame that it becomes visible

dangond-ptc:
> As far as I can tell what will happen is the object's matrices will then be calculated on the frame that it becomes visible

Oh, that's perfect then!

dangond-ptc:
Only problem I can see is that if we ever update three in the future, we might need to re-apply this change

hobinjk-ptc:
Yeah, luckily it's one line, otherwise I'd be much more inclined to do this correctly with a fork and all that nonsense (which would still be its own minor nightmare)

PR #446, opened 2023-05-05
Spaghetti now fades out if it is not the one the user is interacting …

dangond-ptc:
…with

This change makes it so that in cases where multiple people were tracked (ex: the Mercedes demo), clicking on one person's spaghetti hides the other people's spaghettis, making it clearer what can be interacted with (It doesn't make sense to select points across spaghettis and is currently not supported, but leads to confusing behavior if attempted). Unselecting the spaghetti makes all other spaghettis visible again.

### DIFF
--- a/src/humanPose/HumanPoseAnalyzerSettingsUi.js
+++ b/src/humanPose/HumanPoseAnalyzerSettingsUi.js
@@ -112,6 +112,16 @@ export class HumanPoseAnalyzerSettingsUi {
     }
     
     setUpEventListeners() {
+        this.root.addEventListener('pointerdown', (event) => {
+            event.stopPropagation();
+        });
+        this.root.addEventListener('pointermove', (event) => {
+            event.stopPropagation();
+        });
+        this.root.addEventListener('pointerup', (event) => {
+            event.stopPropagation();
+        });
+        
         // Toggle menu minimization when clicking on the header, but only if not dragging
         this.root.querySelector('.hpa-settings-header').addEventListener('mousedown', event => {
             let mouseDownX = event.clientX;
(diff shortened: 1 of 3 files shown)


src/humanPose/spaghetti.js
```diff
@@ -75,16 +83,28 @@ class MeasurementLabel {
 const SpaghettiSelectionState = {
     NONE: {
         onPointerDown: (spaghetti, e) => {
+            if (e.deselectedSpaghetti) {
+                // If another spaghetti was deselected on this click, do nothing
```
hobinjk-ptc, line 87: This might be easier to accomplish using stopPropagation() in the deselection handler

benptc (approved):
LGTM

hobinjk-ptc (approved):
Seems reasonable, tested locally and lgtm overall. Good approach to the pointer down event

src/humanPose/HumanPoseAnalyzerSettingsUi.js
```diff
@@ -112,6 +112,16 @@ export class HumanPoseAnalyzerSettingsUi {
     }
     
     setUpEventListeners() {
+        this.root.addEventListener('pointerdown', (event) => {
+            event.stopPropagation();
+        });
+        this.root.addEventListener('pointermove', (event) => {
+            event.stopPropagation();
```
hobinjk-ptc, line 119: This blanket stopPropagation has the unfortunate effect that if you start a camera interaction (e.g. right click to pivot) then go over the settings ui it first stops rotating due to losing pointermove then doesn't stop the camera interaction since pointerup also gets stopped. I think the more durable solution is to check event.target in the spaghetti's pointer listeners. This would also fix clicking on other places like the top menu bar resetting the spaghetti region

PR #448, opened 2023-05-05
Miscellaneous Analytics Fixes

hobinjk-ptc:
Most notably prevents race condition where a frame opening automatically while minimized would send a setDisplayRegion to the fallback 'none' analytics, causing the loading of history to happen two times in parallel. Additionally fixes the initial display of data after a recording finishes and cleans up conditions where the timeline would be too narrow due to very short recordings.


dangond-ptc (approved):
Looks good to me!

PR #451, opened 2023-05-05
Adds -webkit-user-drag:none to all elements; disables hover on mobile ru-icons

benptc:
It seems that setting user-select:none affects selecting the text on elements, touch-action:none affects scrolling and zooming, and pointer-events:none makes the elements completely ignore touches. None of these is what I need to disable long-pressing on certain images triggering iOS's built-in drag-and-drop feature.

Adding `-webkit-user-drag:none` does the trick, though! Originally I was just going to apply it on the minimized tool icons, but instead I'm applying it everywhere, as I saw that there are all sorts of buttons that have a weird "drag" behavior if it isn't added more globally.

I also disabled the lines that appear when you hover over a recently used tool icon, on mobile, since "hover" only makes sense if you have a mouse, not a touchscreen.

Lastly, I slightly increased the animation time of the lines that go from the recently used tool icons, because before it was essentially imperceptible. It's still very subtle but Valentin wants it to stay subtle.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -61,6 +61,7 @@
     -webkit-transform: matrix3d(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1);
    -webkit-user-select: none;
     -webkit-touch-callout:none;
+    -webkit-user-drag: none;
 }
 
 #GUI {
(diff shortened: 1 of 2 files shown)


ktheo-ptc:
Great that's good to know about drag events!

PR #463, opened 2023-05-10
Fixed errors arising from spaghetti.analytics being null (used for ca…

dangond-ptc:
…mera paths)

### DIFF
--- a/src/humanPose/spaghetti.js
+++ b/src/humanPose/spaghetti.js
@@ -8,8 +8,8 @@ import {AnalyticsColors} from "./AnalyticsColors.js";
 
 const spaghettiListsByAnalyticsFrame = {};
 const activeSpaghettiByAnalyticsFrame = {};
-function updateAllSpaghettiColorsByAnalytics(analytics) {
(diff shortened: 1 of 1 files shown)


src/humanPose/spaghetti.js
```diff
@@ -340,10 +340,10 @@ export class Spaghetti extends THREE.Group {
         this.setupPointerEvents();
         this.addPoints(points);
         
-        if (!spaghettiListsByAnalyticsFrame[this.analytics.frame]) {
-            spaghettiListsByAnalyticsFrame[this.analytics.frame] = [];
+        if (!spaghettiListsByAnalyticsFrame[this.frame]) {
```
ptc-mklaudiny, line 343: It's just not clear to me where this.frame is set in this class.

ptc-mklaudiny (approved):
Tested and I did not run into the issue again.

src/humanPose/spaghetti.js
ptc-mklaudiny: ah yes, I missed it.

PR #473, opened 2023-05-25
Fixes issues with tool opening focusing and blurring on reload, fixes…

dangond-ptc:
… history loading into none Analytics session

This was specifically an issue caused by my last pull request setting the activeFrame to 'none' on blur. In this case, refreshing the page would launch an Analytics session and immediately make it inactive (as intended), but `loadHistory` would always use the active analytics, causing the history to load into the 'none' analytics session and not be closeable/on the timeline.


hobinjk-ptc (approved):
LGTM, closer and closer to having everything be fully compartmentalized!

PR #477, opened 2023-06-01
Somewhat hacky fix for live poses not keeping last active analytics' …

dangond-ptc:
…lens mode when minimizing

This feels more intuitive for the user, who otherwise is wondering why live poses keep switching back to REBA mode as the analytics system starts sending live data to the `none` analytics session.

### DIFF
--- a/src/analytics/index.js
+++ b/src/analytics/index.js
@@ -19,6 +19,11 @@ import {AnalyticsMobile} from './AnalyticsMobile.js'
     const noneFrame = 'none';
     let activeFrame = '';
     let analyticsByFrame = {};
+    
+    function getDefaultAnalytics() {
+        return analyticsByFrame[noneFrame];
+    }
+    exports.getDefaultAnalytics = getDefaultAnalytics;
 
     /**
      * @return {Analytics}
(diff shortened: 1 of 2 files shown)


hobinjk-ptc (approved):

ptc-mklaudiny:
I tested and it works for standard lenses, but there are issues with debug Distinct people lens. When the analytics tool is minimized, the auxiliary skeletons disappear and the fused skeleton changes color. 
The reason why I want to minimize in the first place is to have a clear close view on complex situation with several skeletons which is not obstructed by timeline.

dangond-ptc:
> I tested and it works for standard lenses, but there are issues with debug Distinct people lens. When the analytics tool is minimized, the auxiliary skeletons disappear and the fused skeleton changes color.
> The reason why I want to minimize in the first place is to have a clear close view on complex situation with several skeletons which is not obstructed by timeline.

I can add another fix for this by making the live poses always be sent to render in the 'none' analytics session? That way the colors won't change, since the colors are determined by the order in which they are added to that analytics session. Since the selected lenses are synchronized between the active analytics and the 'none' analytics, I think this should work as we expect?

hobinjk-ptc:
I like the 'none' analytics session being used for all live poses. Generally, I think the best solution would be that each analytics session has no concept of live vs historical internally and the 'none' analytics session is all live poses all the time

PR #480, opened 2023-06-06
Send live poses to default analytics

dangond-ptc:
Live poses now only get rendered by the default analytics sessions rather than the sessions associated with tools. Along with this change, the analytics settings menus are now set up such that they don't include settings for both historical and live data. Spaghetti Lines have also been renamed to Motion Paths in the UI. The default analytics session menu (opened via the menu bar) is now labeled Live Analytics Settings instead of Analytics Settings for clarity as well.

### DIFF
--- a/src/analytics/index.js
+++ b/src/analytics/index.js
@@ -93,7 +93,10 @@ import {AnalyticsMobile} from './AnalyticsMobile.js'
                 analyticsByFrame[msgData.frame] = makeAnalytics(msgData.frame);
             }
             if (activeFrame !== msgData.frame) {
-                getActiveAnalytics().blur();
+                const activeAnalytics = getActiveAnalytics();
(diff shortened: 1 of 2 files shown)


hobinjk-ptc (approved):

ptc-mklaudiny (commented):
I tested and it behaved as I would expect. I noticed just two things:
- Lens selection in Analytics Settings (related to some Analytics tool) changes lens for live poses as well. But the other checkboxes in settings do not affect live poses.
- Menu item to toggle analytics settings in top menu - checkmark gets out of sync with whether Live analytics settings are displayed.

ptc-mklaudiny (approved):
Thanks, works fine. I did notice that 'auxiliary poses' checkbox in live settings is synced with a tool's settings (not other way around). But it's not too important since it's a debug flag (ultimately it should not be visible in production versions).

PR #481, opened 2023-06-12
Analytics: Timeline and regionCard improvements

hobinjk-ptc:
- Adds an accent color to region cards and draws them on the timeline's second track
- Makes it clearer when a region card is shown on the timeline for editing
- Adds a play/pause button to pose playback
- Other misc. fixes


dangond-ptc (commented):
Nice changes! It's very nice having the color-coding for the steps. It would be nice to have the whole play/pause button disappear rather than just the icons when no region is selected though. Thoughts?

src/analytics/analytics.js
```diff
@@ -319,6 +319,9 @@ export class Analytics {
             regionCard.setLabel('Step ' + this.nextStepNumber);
         }
 
+        let hue = (this.nextStepNumber * 17) % 360;
+        regionCard.setAccentColor(`hsl(${hue}, 100%, 50%)`);
```
ptc-mklaudiny, line 323: Another idea might be to derive the color from ordering on timeline rather than current step counter.  In practice, such recoloring happens even now when you reload the viewer.

ptc-mklaudiny (approved):
Nice UX imporvements. I appreciate in particular stopping playback and looking at the pose from any perspective.

ptc-mklaudiny:
> Nice changes! It's very nice having the color-coding for the steps. It would be nice to have the whole play/pause button disappear rather than just the icons when no region is selected though. Thoughts?

It did not particularly bother me.

PR #485, opened 2023-06-22
Support optional Vite-based compilation

hobinjk-ptc:
Requires no changes to run locally (e.g. after applying this PR it's expected that nothing should break)

Most notably this makes onInit block until all addon scripts have finished loading which could be a big problem but also single-handedly fixes all of the load order issues we have ever had forever.

- [x] add notes about how to compile in README


src/device/onLoad.js
```diff
@@ -77,7 +77,7 @@ realityEditor.device.onload = function () {
 
     // Add-ons may need to modify globals or do other far-reaching changes that
     // other services will need to pick up in their initializations
-    realityEditor.addons.onInit();
+    await realityEditor.addons.onInit();
```
benptc, line 80: Good addition!

src/addons/index.js
```diff
@@ -1,21 +1,40 @@
...
             if (source.startsWith('/')) {
               source = '.' + source;
             }
             scriptNode.src = source;
             scriptNode.type = 'module';
             document.head.appendChild(scriptNode);
-        }
+            return loe;
+        });
+        return Promise.all(loePromises);
```
benptc, line 37: Nice way to determine that they've all loaded!

index.html
```diff
@@ -87,100 +89,100 @@
...
-    <script src="src/gui/utilities.js"></script>
-    <script src="src/gui/buttons.js"></script>
-    <script src="src/gui/menus.js"></script>
-    <script src="src/gui/pocket.js"></script>
-    <script src="src/gui/settings.js"></script>
-    <script src="src/gui/modal.js"></script>
-    <script src="src/gui/dropdown.js"></script>
-    <script src="src/gui/glRenderer.js"></script>
-    <script src="src/gui/navigation.js"></script>
+    <script src="src/device/index.js" type="module"></script>
```
benptc, line 111: nice to have everything as modules, now!

benptc (approved):
Tested locally and LGTM!

PR #487, opened 2023-06-26
Pinching triggers remote operator "VR mode" to load within the AR app

benptc:
Also adds another blue overlay circle that goes to the second finger. And adds API support for setFull2D which is useful for loading third-party-embedded 2D tools

Requires updates to pop-up-onboarding and remote-operator-addons, coming soon


.eslintrc.js
```diff
@@ -42,7 +42,8 @@ module.exports = {
         "hull": "writable",
         "io": "writable",
         "objects": "writable",
-        "overlayDiv": "writable",
+      "overlayDiv": "writable",
```
hobinjk-ptc, line 45: nit: indentation

hobinjk-ptc (approved):
Overall LGTM, some minor lint issues (overlayDiv2 should be part of the huge `exported` block at the top of states.js)

PR #500, opened 2023-07-25
Secret support

dangond-ptc:


dangond-ptc:
https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/855

src/oauth/index.js
```diff
@@ -5,8 +5,20 @@ import { loadToken } from './tokens.js';
...
+            });
+            // The edge server that will handle the processing of the auth code to receive an auth token
+            const edgeServer = `http://${object.ip}:${object.port}`; // TODO: more resilient way to do this, not sure if toolbox edge allows connecting via ip:port
+
+            let frameName = frame;
+            realityEditor.forEachFrameInAllObjects((objectKey, frameKey) => { // TODO: make postMessages automatically pass the originating frame to callbacks, make things cleaner here
+                if (frameKey === frame) {
+                    frameName = realityEditor.getFrame(objectKey,frameKey).src;
+                }
+            });
```
benptc, line 20: Could this be simplified since you already find the `object` above, to something like `object.frames[frame].src`, rather than looping over all objects/frames?

benptc (approved):
Looks like it should work – one small comment inline

hobinjk-ptc (approved):

dangond-ptc:
Made more changes to properly support proxying through toolboxedge.net and alpha.toolboxedge.net

src/oauth/index.js
```diff
@@ -3,10 +3,22 @@ createNameSpace("realityEditor.oauth");
 import { loadToken } from './tokens.js';
 
 (function(exports) {
+    function getToolboxEdgeBasePath() {
+        const windowPath = window.location.pathname;
+        return '/' + windowPath.split('/').slice(2,6).join('/') + '/i/whud7837yhd'; // => (/stable dropped when connecting to edge-server rather than ui) /n/networkId/s/networkSecret
```
hobinjk-ptc, line 8: What is the 6 accomplishing here?

dangond-ptc: Limiting the base path to `/n/networkId/s/networkSecret` and not anything further beyond that. I don't think it's actually necessary, but it was mostly general paranoia about accidentally having the relative path of the script included in the URL. I can change it to `.slice(2)` if you think that would be cleaner.

hobinjk-ptc: I tend to be against any and all magic numbers but it's certainly a better solution than fully parsing the /n/{}/s/{} url structure to manually remake it. I'm okay with it

src/oauth/index.js
```diff
@@ -3,10 +3,22 @@ createNameSpace("realityEditor.oauth");
...
     function initService() {
         realityEditor.network.addPostMessageHandler('getOAuthToken', (msgData) => {
-            const { frame, urls, clientId, clientSecret } = msgData;
-            loadToken(frame, urls, clientId, clientSecret).then(token => {
+            const { frame, authorizationUrl, clientId } = msgData;
+            const object = Object.values(realityEditor.objects).find(obj => {
+                return Object.keys(obj.frames).includes(frame);
+            });
+            // The edge server that will handle the processing of the auth code to receive an auth token
+            const edgeServer = window.location.origin.includes('toolboxedge') ? window.location.origin + getToolboxEdgeBasePath() : `http://${object.ip}:${object.port}`;
```
hobinjk-ptc, line 18: It would be good for all of these window.location.origin.includes() to be refactored out to a single IS_CLOUD variable to make it easier for the Atlas-hosted version of this to rewrite toolboxedge to something else

hobinjk-ptc (approved):
LGTM! One comment inline

src/oauth/redirect.js
```diff
@@ -1,47 +1,66 @@
...
     const code = parameters.get('code');
     const error = parameters.get('error');
     const nonce = parameters.get('state');
+    const state = JSON.parse(localStorage.getItem('activeOAuthState')) || {};
     if (localStorage.getItem('activeOAuthNonce') !== nonce) {
-        // TODO: error handling
-        console.log(`${localStorage.getItem('activeOAuthNonce')} does not match ${nonce}`);
-        localStorage.removeItem('activeOAuthNonce');
-        localStorage.removeItem('activeOAuthState');
+        handleError(`${localStorage.getItem('activeOAuthNonce')} does not match ${nonce}`, state.toolboxUrl);
```
benptc, line 32: Great to see you've added the error handling!

benptc (approved):
LGTM

PR #502, opened 2023-07-26
load rvl earlier and add error handling around the initService calls …

benptc:
…in onLoad.js to fix issue where vuforia engine might never initialize. Fixes issue if scripts load in the wrong order.


hobinjk-ptc (approved):
Might be good to have a case for if realityEditor.gui.modal.showScreenTopNotification fails with an old fashioned call to `alert`

benptc:
> Might be good to have a case for if realityEditor.gui.modal.showScreenTopNotification fails with an old fashioned call to `alert`

Good idea! Added that

PR #513, opened 2023-08-17
VideoPlayback extends Followable class; can switch to first person video

benptc:
- Sets up videoPlayback to work with the new cameraFollowCoordinator in the remote operator
- Renders a camera mesh at the position that the videoPlayback was recorded from 
- Ability to switch between volumetric and first-person shaders for video playback


hobinjk-ptc (approved):
Tested locally and now LGTM

PR #514, opened 2023-08-21
Avatar sync update

ktheo-ptc:
- avatar sync bug fix


benptc (approved):
Great work tracking this one down! This fix has escaped me for so many months

PR #517, opened 2023-08-25
increase renderOrder of spatial cursors so they don't z-fight with 2d virtualizer video

benptc:
This has been bugging me for awhile – following a video in first person mode would cause the spatial cursor and the virtualizer video to z-fight since they both turn off depthTest and had the same renderOrder.

Before update:
broken-render-order-spatial-cursors 2023-08-25 11_53_48

After update:
fixed-render-order-spatial-cursors 2023-08-25 11_54_50


Steve-KX-RL (approved):
This is very nice to have!

hobinjk-ptc (approved):
Nice fix!

PR #521, opened 2023-09-13
Profiling UI v1

benptc:
Adds the beginnings of the settings UI for an "Activity Monitor" / "Profiler".
Adds handlers for new APIs that tools can use to log messages / timings / counts to this UI.
Adds a first example metric, which measures the time between setting the camera position, and the time when the updated matrices have been sent into all tools that are subscribed to matrices.

profiler-ui-v1


dangond-ptc (approved):
Looks good to me!

PR #522, opened 2023-09-14
dont download all icon.gif files if UI is overridden (e.g. with pop-up-addon)

benptc:
Decreases the number of network requests the system makes when it starts up – since we don't show the old pocket anymore, we probably don't need to download all of the icon.gif files for the tools we never see. As a side effect, eliminates some error messages from the console that always show up on launch (depending on which addons/tools you have)

Screenshot 2023-09-14 at 10 08 49 AM


dangond-ptc (approved):
Good to have this!

hobinjk-ptc:
Great to have!

hobinjk-ptc (approved):

PR #523, opened 2023-09-14
implements new sendCoordinateSystems API and adds a few optimizations

benptc:
Implements the APIs added in https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/891

Also moves the floorOffset to always be at the lowest point in the mesh, and adds a couple minor optimizations to the camera update loop.

### DIFF
--- a/src/gui/threejsScene.js
+++ b/src/gui/threejsScene.js
@@ -611,7 +611,6 @@ import { ViewFrustum, frustumVertexShader, frustumFragmentShader, MAX_VIEW_FRUST
         return groundPlaneCollider;
     }
 
-    // TODO: implement this without relying on worldOcclusionObject, so that it can be used as a backup method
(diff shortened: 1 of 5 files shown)


dangond-ptc (approved):

src/gui/threejsScene.js
Steve-KX-RL, line 1: Would it make more sense to move all the matrix-related functions into a new file, eg. matrix.js, to separate them from all the three.js specific things? Not an urgent request but could be future-proof.

Steve-KX-RL (approved):
Today we rejoice the new APIs on the block👼

hobinjk-ptc (approved):

src/gui/threejsScene.js
benptc: Yeah that's a good idea. I think I'll merge this as-is for now, so that the APIs are available, but you're right that this could use a refactor.

PR #525, opened 2023-09-20
hide other tools and icons in the background when a full2d tool is open 

benptc:
I think the UI looks cleaner if they are hidden while the background is blurred, since they don't blur and sometimes interfere with the focused tool's content.

Example before:
Screenshot 2023-09-20 at 12 21 21 PM

Example after:
Screenshot 2023-09-20 at 12 20 33 PM

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -840,6 +840,11 @@ body {
     pointer-events: none;
 }
 
+.hiddenFrameContainerByFull2D {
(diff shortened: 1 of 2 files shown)


ktheo-ptc (approved):
Great update!

hobinjk-ptc (approved):
Excellent to have this!

PR #527, opened 2023-09-28
Fix bug in previous implementation of "hidden by blurred background"

benptc:
Improvement based on @ptc-mklaudiny's observation that the time-series graph stops ticking forward while the blurred background is hiding it in the previous implementation (https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/525).

A similar side effect seems to be that a proxied gl context doesn't render while display:none is applied to the corresponding iframe, so I didn't need to specifically hide the glcanvas in the previous implementation.

This solution uses visibility:hidden on the iframes, but as a result also needs to apply visibility:hidden to the glcanvas.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -840,11 +840,6 @@ body {
     pointer-events: none;
 }
 
-.hiddenFrameContainerByFull2D {
-    display: none;
-    pointer-events: none;
-}
-
 .visibleFrame {
     visibility: visible;
     /*opacity: 1.0;*/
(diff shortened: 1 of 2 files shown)


hobinjk-ptc (approved):
Interesting, nice improvement! Good ol' requestAnimationFrame

ptc-mklaudiny (approved):
Thanks for fixing this issue.

PR #528, opened 2023-10-03
Loosen comparison of target ip if one ip is localhost

hobinjk-ptc:
Required to fix https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/909

### DIFF
--- a/src/app/targetDownloader.js
+++ b/src/app/targetDownloader.js
@@ -696,7 +696,8 @@ createNameSpace("realityEditor.app.targetDownloader");
         for (var objectKey in objects) {
             if (!objects.hasOwnProperty(objectKey)) continue;
             const object = realityEditor.getObject(objectKey);
-            if (object.ip === ip && object.name === objectName) {
+            const ipMatches = object.ip === ip || object.ip === 'localhost' || ip === 'localhost';
+            if (ipMatches && object.name === objectName) {
                 return objectKey;
             }
         }
--- a/src/network/realtime.js
+++ b/src/network/realtime.js
@@ -665,7 +665,8 @@ createNameSpace("realityEditor.network.realtime");
         if (socketIP.includes(':8081')) {
             ioObject = window._oldIo.connect(socketIP);
         } else {
-            ioObject = io.connect(socketIP);
+            let tsIo = new ToolSocket.Io();
+            ioObject = tsIo.connect(socketIP);
         }
         if (DEBUG) {
             console.log('createSocketInSet', setName, socketIP, ioObject);


src/network/discovery.js
```diff
@@ -135,7 +135,9 @@ createNameSpace("realityEditor.network.discovery");
 
     exports.setPrimaryWorld = (ip, id) => {
         primaryWorld = {
-            ip: ip,
```
dangond-ptc, line 138: Is there nothing that uses the ip value here?

hobinjk-ptc: The code does use the ip value if present (in src/network/index.js). However, I think that setting the ip address is what causes joining an existing scan from the metaverse manager to fail.

More precisely, the following:
device A and device B are logged into realitydemo@ptc.com in the metaverse manager
device A creates a new scan
device B sees the scan in the metaverse manager, clicks Join Metaverse
right now: device B doesn't localize or download the area target
with this change (expected, can't test locally right now): device B successfully sees that device A is hosting the expected world id

dangond-ptc (approved):

benptc:
Seems reasonable. I just tested with multiple devices and this is working!

1. Works perfectly if you sign into the same account on multiple devices, and tap on "Enter Metaverse" for the metaverse you scanned. The only problem is you need to remember which device actually scanned the world in the first place, as you need to join on that device first. Not sure what the long-term solution for this would be – some visual feedback in the manager that distinguishes between metaverses stored on this device vs ones scanned on other devices? Or to eventually store the target data in the cloud so that you could start the session on any device logged into this account.

2. I was also able to join via the "Join Existing Session" menu – however, there is an issue if you have many metaverses scanned on your device where the UI in the join menu has a button for each world ID/IP pair (but only shows the IP) so it's a bit of a game of luck to pick the right button that corresponds with the world that the other device is localized within. Ideally the other iOS device's server wouldn't advertise the worlds that aren't actively being used (perhaps deactivating the object?). Or at least we would have better UI in this menu to help you pick the right world.

join-existing-metaverse-list-of-too-many

hobinjk-ptc:
Thanks for testing! Yeah, the current lack of distinction is definitely a stumbling block for users. I agree that the eventual proper solution is to download the target from the cloud.

Ideally the server would only activate the beat sending for activated-in-the-ui world objects but I'm not sure how implementable that is

src/network/discovery.js
hobinjk-ptc: I decided to drop this blunt approach in favor of selectively not providing `ip` in callers of this function

PR #533, opened 2023-10-11
Take spatial snapshots on iOS device; refactor SpatialSnapshotCoordinator out of CameraVisCoordinator; implement capture APIs

benptc:
- moves Shaders.js, VisualDiff.js, and CameraVisPatch.js to userinterface, so that patches can be added in AR
- creates new SpatialPatchCoordinator class, with a subset of functionality from the CameraVisCoordinator
- implements `spatialInterface.captureSpatialSnapshot` tool API (https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/913)
- implements swift `get3dSnapshot` API (https://github.com/ptcrealitylab/vuforia-spatial-toolbox-ios-swift/pull/495)
- Adds a bit more logic to the SOLID shader (adds viewAngleSimilarity and viewPositionSimilarity uniforms) to make photo smoother when viewing from similar perspective that it was captured from
- adjusts the coordinate system that spatial patches are positioned within to make the math easier to calibrate between remote operator and iOS views

Corresponding remote operator PR coming momentarily...

### DIFF
--- a/index.html
+++ b/index.html
@@ -170,6 +170,9 @@
     <script src="src/gui/ar/areaCreator.js" type="module"></script>
     <script src="src/gui/ar/videoPlayback.js" type="module"></script>
     <script src="src/gui/ar/Followable.js" type="module"></script>
+    <script src="src/gui/ar/Shaders.js" type="module"></script>
+    <script src="src/gui/ar/VisualDiff.js" type="module"></script>
+    <script src="src/gui/ar/CameraVisPatch.js" type="module"></script>
 
     <script src="src/gui/spatial/index.js" type="module"></script>
     <script src="src/gui/spatial/timeRecorder.js" type="module"></script>
--- a/index.html
+++ b/index.html
@@ -195,9 +195,6 @@
     <script src="src/spatialCursor/shader/colorCursorFragmentShader.js" type="module"></script>
     <script src="src/spatialCursor/shader/vertexShader.js" type="module"></script>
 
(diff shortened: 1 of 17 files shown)


index.html
```diff
@@ -195,6 +195,11 @@
     <script src="src/spatialCursor/shader/colorCursorFragmentShader.js" type="module"></script>
     <script src="src/spatialCursor/shader/vertexShader.js" type="module"></script>
 
+    <script src="src/spatialCapture/Shaders.js" type="module"></script>
+    <script src="src/spatialCapture/VisualDiff.js" type="module"></script>
+    <script src="src/spatialCapture/CameraVisPatch.js" type="module"></script>
+    <script src="src/spatialCapture/SpatialPatchCoordinator.js" type="module"></script>
```
hobinjk-ptc, line 201: Since SpatialPatchCooordinator imports CameraVisPatch which imports the other two files only the SpatialPatchCoordinator <script> tag is needed here

src/index.js
```diff
@@ -149,6 +149,12 @@ window.realityEditor = {
             vertexShader: {},
         }
     },
+    spatialCapture: {
+        CameraVisPatch: {},
```
hobinjk-ptc, line 153: I don't see anything filling out these empty objects. It looks more like this should be `spatialCapture: {}` for the later `realityEditor.spatialCapture.spatialPatchCoordinator = new SpatialPatchCoordinator();`

hobinjk-ptc (approved):

PR #535, opened 2023-10-13
Value Add/Waste Time functionality for analytics

dangond-ptc:
Apologies for the massive mono-commit! This brings in the ability to mark parts of the analytics timeline as being Value Add or Waste Time. Right now, the only visualization of this is as regions on the timeline, but I'm looking into where a good spot to put a summary would be. Depends on https://github.com/ptcrealitylab/vuforia-spatial-core-addon/pull/289 and https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/924


dangond-ptc:
Also, this has the unfortunate side-effect of not being backwards-compatible with existing analytics sessions. I'll also add in a commit that can change the previous data format into the new one.

src/utilities/Timeline.js
```diff
@@ -0,0 +1,361 @@
+/**
+ * A Region represents a chunk of time [startTime, endTime) with a given value.
+ */
+export class Region {
```
benptc, line 4: I like how nicely structured the Region and Timeline classes are – nice work!

src/utilities/Timeline.test.js
```diff
@@ -0,0 +1,346 @@
+import { expect, test, describe } from 'vitest';
+import { Region, Timeline } from "./Timeline";
+
+const region = new Region(0, 5, null);
+
+describe('region', () => {
```
benptc, line 6: Nice to see the tests here, too!

hobinjk-ptc (approved):
Very nice API design! Even with tests! Incredible

benptc (approved):
Looks good!

PR #538, opened 2023-10-19
Fixes OAuth redirecting to new scan page on mobile, now redirects to …

dangond-ptc:
…join scan page


benptc (approved):
I haven't tested locally, but code LGTM based on my knowledge of how the newScan/joinScan URLs were working

PR #541, opened 2023-10-31
Render my avatar pointers

Steve-KX-RL:


benptc (approved):
I added some updates so that multiple lines can be drawn on the screen at the same time (previously the clearLinkCanvas would delete the other lines every time it draws a new one), and to draw the line with a thicker width, and to draw from a default position if it can't find the avatar icon for that avatar (e.g. if there are too many avatars and this one is contained within the "+2" icon)

benptc:
Thanks for your help with this, Steve!

Steve-KX-RL:
> I added some updates so that multiple lines can be drawn on the screen at the same time (previously the clearLinkCanvas would delete the other lines every time it draws a new one), and to draw the line with a thicker width, and to draw from a default position if it can't find the avatar icon for that avatar (e.g. if there are too many avatars and this one is contained within the "+2" icon)

Awesome, thanks Ben!

PR #542, opened 2023-11-01
fix the restoration of the virtualizer visibility after you press P to clone patches

benptc:
This should fix the issue you and Valentin discovered – can you test? Feel free to merge it if it works


Steve-KX-RL:
> This should fix the issue you and Valentin discovered – can you test? Feel free to merge it if it works

I tested it locally and it's working --- thx Ben!

PR #544, opened 2023-11-06
Miscellaneous threejsScene fixes and improvements

hobinjk-ptc:
Adds back the fancy loading animation (with frustum culling still disabled)
Fixes issue where on larger scans the white mesh areas would show up


benptc (approved):
Nice improvements. IIRC the areaTargetMaterialWithTextureAndHeight is still a tiny bit of a performance hit on older laptops because it's doing some unnecessary computations even with frustum culling turned off (such as adding barycentric coordinates to all vertices to prepare for rendering culled areas as wireframes), but I think it should be fine.

hobinjk-ptc:
Sounds good, something to keep in mind for sure

PR #549, opened 2023-11-17
Renamed analytics to motionStudy to work around adblockers

dangond-ptc:


dangond-ptc:
Everything other than strings, css classes and ids, and tool APIs has been renamed

benptc (approved):
Thanks for the refactor! Tested locally (not creating a new analytics tool, but at least using an existing recorded pose), and all seems to work

PR #561, opened 2023-12-06
Hides non-fullscreen tools/icons when their MVP matrix has negative matrix[14]

benptc:
This has nagged at me for awhile but I never tracked down what was causing it.

Prevents icons / CSS-3D transformed tools from rendering when you zoom past them – the way we manipulate the matrix in the render loop currently causes them to show up upside-down. Sets display:none when conditions are met. Only does this on non-fullscreen tools, which ensures that 3D content or an active envelope doesn't hide when you move the camera.

Note: I added a new CSS class `elementBehindCamera` to tag these with instead of re-using `outsideOfViewport` because outsideOfViewport has some complicated logic attached that involves reloading the iframe when it moves in and out (as calculated by `realityEditor.gui.ar.positioning.canUnload`) which is disabled on desktop, and I don't think we need to unload/reload icons each time you move the camera past them.

Old version:
tools-rendering-upside-down

New version:
fixed-tools-rendering-upside-down

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -885,6 +885,10 @@ body {
     opacity: 0.7;
 }
 
+.elementBehindCamera {
+    display: none;
+}
+
 .outsideOfViewport {
     display: none;
 }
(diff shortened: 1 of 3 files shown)


Steve-KX-RL (approved):
I've never discovered this issue before. Good catch!

PR #557, opened 2023-12-04
Improved transformation for billboarding the icons

benptc:
Old icon billboarding looked like:
old-billboarding-demo

New icon billboarding looks like:
new-billboarding-demo

PR #562, opened 2023-12-06
revert to old billboarding on AR devices; use new method on desktop

benptc:
In my enthusiasm to improve the billboarding for remote operators (https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/pull/557) I forgot to test how it impacts rendering in AR, and now I've realized that the new method I used (basically just setting rotation to identity) doesn't work on AR because when you rotate the phone (like a steering wheel) the icons rotate with it. I've tried various maths for getting the perfect transform in AR but can't quite get it to avoid singularity behavior when you view from the top. Something something quaternions but I don't have the headspace for that right now.

This reverts to the old method on AR but keeps the new implementation for desktop.


hobinjk-ptc (approved):
Oh, good point!

PR #567, opened 2023-12-08
Don't load world object on proxy

hobinjk-ptc:
App update not necessary, only a change on toolboxedge.net


benptc (approved):
Looks good to me, assuming we continue to load the userinterface locally on the phone instead of eventually loading it from the proxy too (but if we want to support offline mode then that would never happen)

PR #571, opened 2023-12-11
Remove uses of oldIo; Expose variables necessary for other rip-out-ws branches

hobinjk-ptc:


benptc (approved):
Nice to get rid of that

PR #573, opened 2023-12-12
Misc error guards and whitespace cleanup

benptc:
I added `if (!realityEditor.motionStudy.getAllHumanPoseAnalyzers) return;` because I think `getAllHumanPoseAnalyzers` is undefined and this function is able to be triggered by pressing the H keyboard shortcut. Not sure what the intended functionality is.

### DIFF
--- a/src/humanPose/draw.js
+++ b/src/humanPose/draw.js
@@ -353,18 +353,6 @@ function toggleAnalyzerSettingsUI() {
 }
 
 /**
- * Sets the visibility of the human poses
- * @param {boolean} visible - whether to show or not
- */
-function setHumanPosesVisible(visible) {
-    if (!realityEditor.motionStudy.getAllHumanPoseAnalyzers) return;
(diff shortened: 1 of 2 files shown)


src/humanPose/draw.js
```diff
@@ -357,6 +357,7 @@ function toggleAnalyzerSettingsUI() {
  * @param {boolean} visible - whether to show or not
  */
 function setHumanPosesVisible(visible) {
+    if (!realityEditor.motionStudy.getAllHumanPoseAnalyzers) return;
```
hobinjk-ptc, line 360: This whole function (and corresponding shortcut) can be removed. There's not much of a reason to provide a way to hide human poses when we also have a way to disable human pose tracking

hobinjk-ptc (approved):
LGTM, death to trailing whitespace!

src/humanPose/draw.js
benptc: Ok, I'll remove those instead

PR #582, opened 2024-01-04
updates receiveMatricesFromAR to be compatible with targetId tracking method

benptc:
New app version reports tracked targets using their targetId, rather than objectId. This method is still backwards compatible with old app versions or object versions which don't have a targetId.

Goes with https://github.com/ptcrealitylab/vuforia-spatial-toolbox-ios-swift/pull/555

### DIFF
--- a/src/app/callbacks.js
+++ b/src/app/callbacks.js
@@ -217,7 +217,15 @@ createNameSpace('realityEditor.app.callbacks');
      * Callback for realityEditor.app.getMatrixStream
      * Gets triggered ~60FPS when the AR SDK sends us a new set of modelView matrices for currently visible objects
      * Stores those matrices in the draw module to be rendered in the next draw frame
-     * @param {Object.<string, Array.<number>>} visibleTargets
+     * @param {Object.<string, {matrix: number[], targetName: string}>} visibleTargets.current - (App versions starting 2024)
+     *   - Each key is a unique target ID, used internally by Vuforia. The value is an object with:
+     *   - `matrix`: Length-16 transformation matrix representing the target relative to Vuforia's (0,0,0)
+     *   - `targetName`: The name of the target. In old versions, this is the objectId. In current versions it can be anything.
+     * @param {Object.<string, number[]>} visibleTargets.deprecated - (Deprecated app versions - prior to 2024)
+     *   - Each key is the target name, not the target ID, and the value is the matrix (array of 16 numbers)
+     *   - In this version, the target must be generated using the corresponding objectId as its target name
+     *
+     * Note: The deprecated format is maintained for backward compatibility.
      */
     function receiveMatricesFromAR(visibleTargets) {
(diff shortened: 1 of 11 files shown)


src/app/callbacks.js
```diff
@@ -224,28 +217,37 @@ createNameSpace('realityEditor.app.callbacks');
      * Callback for realityEditor.app.getMatrixStream
      * Gets triggered ~60FPS when the AR SDK sends us a new set of modelView matrices for currently visible objects
      * Stores those matrices in the draw module to be rendered in the next draw frame
-     * @param {Object.<string, Array.<number>>} visibleObjects
+     * @param {Object.<string, Array.<number>>} visibleTargets
```
ptc-mklaudiny, line 220: The type could be documented for current version

src/app/callbacks.js
```diff
@@ -224,28 +217,37 @@ createNameSpace('realityEditor.app.callbacks');
...
+            } else {
+                // find the object whose targetId matches the key, and store the .matrix under the objectId key
+                let matchingObjectId = Object.keys(objects).find(objectKey => {
+                    return objects[objectKey].targetId === key;
+                });
+                if (matchingObjectId) {
+                    visibleObjects[matchingObjectId] = visibleTargets[key].matrix;
+                } else {
+                    // for old objects that weren't generated with a targetId, try to find them by targetName
+                    visibleObjects[visibleTargets[key].targetName] = visibleTargets[key].matrix;
```
ptc-mklaudiny, line 248: So status and status info of targets was never propagated further into the JS side of system ?

ptc-mklaudiny (approved):
Looks fine.

src/app/callbacks.js
benptc: Currently there was nothing making use of the status / status info, so we just passed the matrix for simplicity/efficiency. If you can think of a way to make use of that data then I can change it to include the status / status info.

ptc-mklaudiny: I was thinking about notifications about potentially inaccurate camera tracking. This should be a combination of status from device tracking and area target tracking (coming through visibleTargets here). But that's for another PR.

benptc: That makes sense! We do have some of this functionality, as the status and statusInfo are included in the getCameraMatrixStream callback https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/70a35bd58659c4186a75f28effe69ae6bd78cd77/src/app/callbacks.js#L425 and handled in the src/device/tracking.js https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/70a35bd58659c4186a75f28effe69ae6bd78cd77/src/device/tracking.js#L91

So the device tracking side of things is already accounted for, but I agree it would be good to incorporate the status from the area target tracking, too.

hobinjk-ptc (approved):

PR #596, opened 2024-01-22
No local world object loading on toolboxedge.net

hobinjk-ptc:


benptc (approved):
Seems reasonable / will reduce confusion 👍

PR #619, opened 2024-01-31
app.setPause and app.setResume can take a completionHandler callback

benptc:
New promise versions of setPause and setResume that resolve when pause/resume is complete, with success status.

These can be used to ensure that the app has finished pausing/resuming AR before continuing, since the regular app function calls aren't guaranteed to be synchronous.


ptc-mklaudiny (approved):
LGTM

ptc-mklaudiny (approved):
LGTM

Issue #614, opened 2024-01-29
Allow you to edit your own username, if you've joined as an anonymous user (anyone who clicks on the invite link generally joins as Anonymous)


benptc:

benptc:
Feature has been added!

PR #620, opened 2024-01-31
Click on icon to edit your username

benptc:
Brings up a modal to edit your username. Instantly propagates to other clients when you hit submit. Tested working on iOS app too. Saves name in browser localStorage and uses as default when you refresh the page if you're an anonymous user. Your signed-in username in the Manager overrides the saved value if you're signed in (the next time you refresh the page, but you can still edit for this session).

Edit Username v1

Fixes Issue #614

### DIFF
--- a/src/gui/modal.js
+++ b/src/gui/modal.js
@@ -72,6 +72,7 @@ createNameSpace("realityEditor.gui.modal");
      * @param {string} submitButtonText
      * @param {function} onCancelCallback
      * @param {function} onSubmitCallback
+     * @param {boolean} useSmallerVersion
      */
     function openClassicModal(headerText, descriptionText, cancelButtonText, submitButtonText, onCancelCallback, onSubmitCallback, useSmallerVersion) {
         // create the instance of the modal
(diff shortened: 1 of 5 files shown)


src/gui/modal.js
```diff
@@ -102,6 +103,65 @@ createNameSpace("realityEditor.gui.modal");
...
+        fade.classList.add('modalVisibleFadeIn');
+        document.body.appendChild(fade);
+
+        // create the container with the header, description, a text input, and a submit and cancel button
+        let container = document.createElement('div');
+        container.classList.add('inputModalCard');
+        // container.classList.add('viewCard', 'center', 'popUpModal');
+        let text = document.createElement('div');
+        text.classList.add('inputModalCardText');
+        text.innerHTML = `<h3>${headerText}</h3>
```
hobinjk-ptc, line 120: It would be better to construct the h3 and p elements using createElement and innerText just in case at some point in the future the headerText or descriptionText is provided from an untrusted source

hobinjk-ptc (approved):

Steve-KX-RL (approved):
Thx for doing this!

PR #627, opened 2024-02-16
Stability Improvements

dangond-ptc:
Toolsocket has been updated to a more recent version, and the userinterface now properly connects to the cloud proxy websocket using `ws` or `wss` as appropriate, allowing for local cloud proxy connections


benptc (approved):
LGTM

PR #635, opened 2024-02-27
Fix for getURL not expecting proxyPort being empty sometimes

dangond-ptc:


benptc (approved):
This should do the trick – thank you!

PR #639, opened 2024-03-06
Remove ACES filmic tone mapping

hobinjk-ptc:
It brings a bit of drama to the scan lighting but we're not in the business of drama


ptc-rdeleeuw (approved):
I think this is for the better. if we scan a car for example we don't want tone mapping change the color. Ideally the camera and screen/headset are calibrated to display the actual color of the paint job on the car. 
Efforts to change the scanned colors to improve the color accuracy of the scanned object should not be applied to the whole scene.

PR #640, opened 2024-03-06
Single layer grid

ptc-rdeleeuw:
simplify the render loop by eliminating the need for layers, rendering the whole scene at once instead of multiple times with different layers

depends on:
https://github.com/ptcrealitylab/vuforia-spatial-remote-operator-addon/pull/158

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ node_modules/
 dist/
 
 coverage/
+/.vscode
--- a/src/gui/ar/groundPlaneRenderer.js
+++ b/src/gui/ar/groundPlaneRenderer.js
@@ -90,7 +90,6 @@ createNameSpace("realityEditor.gui.ar.groundPlaneRenderer");
             const colorGrid = new THREE.Color(realityEditor.device.environment.variables.groundWireframeColor);
             gridHelper = realityEditor.gui.threejsScene.createInfiniteGridHelper(gridSquareSizeInMm, gridRegionSizeInMm, 0.075, colorGrid, maxVisibilityDistanceInMm);
             gridHelper.name = 'groundPlaneVisualizer';
-            gridHelper.layers.set(2);
             realityEditor.gui.threejsScene.addToScene(gridHelper, {occluded: true});
         }
 
--- a/src/gui/threejsScene.js
+++ b/src/gui/threejsScene.js
@@ -30,7 +30,6 @@ import { MapShaderSettingsUI } from "../measure/mapShaderSettingsUI.js";
     let distanceRaycastVector = new THREE.Vector3();
     let distanceRaycastResultPosition = new THREE.Vector3();
     let originBoxes = {};
-    let hasGltfScene = false;
     let allMeshes = [];
     let isHeightMapOn = false;
     let isSteepnessMapOn = false;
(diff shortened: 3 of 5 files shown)


Steve-KX-RL (commented):
LGTM. Is there any reason why we set up layers in the past?

hobinjk-ptc (changes requested):
Very promising technique for resolving the issues we currently use layers for, nice work! Unfortunately, the use of layers in rendering is not only limited to the ground plane grid. We also use them to enable the point cloud rendering to draw above the scan (there used to be some more complex techniques in use as well (the stenciling work) but for our current visuals just drawing above would be sufficient). See attached screenshots for a comparison:
Screenshot 2024-03-06 at 10 52 49 AM

Screenshot 2024-03-06 at 10 52 31 AM

ptc-rdeleeuw:
The main reason for this change is that webxr mode hardcodes layer 1 to the left eye and layer 2 to the right eye, layer 0 (default) renders to both eyes. Where can I find this point cloud rendering code? I would like to make that layer-less as well.

hobinjk-ptc:
For sure! I agree that that's a good goal.

https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/main/src/spatialCapture/Shaders.js is the source. Some examples of how the exports of this file are used: https://github.com/ptcrealitylab/vuforia-spatial-remote-operator-addon/blob/main/content_scripts/CameraVis.js is the real-time live streaming and https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/main/src/spatialCapture/CameraVisPatch.js is the static captures.

benptc (approved):
This seems like a good optimization. LGTM so far.

@Steve-KX-RL has also modified the TransformControls to work with our layers (not really used at the moment but will likely be used in future for moving 3D content around more easily). Steve, would the TransformControls still work without the layers?
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/single_layer_grid/thirdPartyCode/three/TransformControls.js

Steve-KX-RL:
> This seems like a good optimization. LGTM so far.
> 
> @Steve-KX-RL has also modified the TransformControls to work with our layers (not really used at the moment but will likely be used in future for moving 3D content around more easily). Steve, would the TransformControls still work without the layers? https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/single_layer_grid/thirdPartyCode/three/TransformControls.js

It would need some minor tweaking but still works. We can always re-enable some optional layers if we only want to pick specific stuff.

ptc-rdeleeuw:
For spatial photo/video I've used the render order to fix the issue.
- first we render the scanned background
- then we disable the depth test, and render the pointcloud (depth only), this will overwrite the background depth values, ignoring them during the rest of the render
- finally we render all the elements in the scene (without the back ground) including the pointclouds for color and depth checking

ptc-rdeleeuw:
Layer 10 will contain additional gltf models in the future but is not currently in use so we can delay solving that issue until work on that feature is continued

Steve-KX-RL (approved):
LGTM, thx Rik!

hobinjk-ptc (approved):

PR #644, opened 2024-03-12
Update notifications

ktheo-ptc:
- updated status bar notification style


src/gui/modal.js
```diff
@@ -343,20 +343,14 @@ createNameSpace("realityEditor.gui.modal");
     function showScreenTopNotification(message, timeMs = 3000) {
         let notificationUI = document.createElement('div');
         notificationUI.classList.add('statusBar');
-        if (realityEditor.device.environment.variables.layoutUIForPortrait) {
```
benptc, line 346: Nice to get rid of this and use the media query instead!

benptc (approved):
Nice improvement!

Issue #645, opened 2024-03-13
Blend Mode bug on iOS 17.4 / iPhone 15 Pro Max


benptc:
First seen on March 11, 2024.
Happens on TestFlight build 23, 24, and 25 (older builds weren't available to test at the time).
Happens on iPhone 15 Pro Max running iOS 17.4
Doesn't happen on iPad Pro running iPad OS 17.3.1
Doesn't happen on iPhone 13 Pro Max running iOS 16.6

Most easily reproduced by swiping down to see notifications and then returning to the app, then tapping on the screen. Pinching to see VR mode also might trigger it. Usually toggles between dark and normal visuals when user taps on the screen. Drawing tool lines also have aliased black outline, and kebab and addTool menus have dark background. End Session modal may have a darker-than-usual shadow. Doesn't always reproduce.

<img width="300" alt="Screenshot 2024-03-13 at 9 47 59 AM" src="https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/assets/32580292/7e18b5e5-c052-416f-8e3a-145dbd4cbe14">

<img width="300" alt="bug-main-screen-iphone-15-ios-17-4" src="https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/assets/32580292/bcb20a7f-979a-4a72-b456-864f06949a87">

<img width="300" alt="bug-loading-screen-iphone-15-ios-17-4" src="https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/assets/32580292/b20449b9-94da-48d4-b74c-cd62466fcc67">

benptc:
Seems to be fixed by deleting or hiding the 'ru-canvas' from the dom. Unclear why this would be causing this issue.

benptc:
Resolved by clearing canvas; it seems that default behavior of canvases changed in iOS 17.4

PR #646, opened 2024-03-13
hide recently-used canvas on AR app, since it was causing weird blend mode bug

benptc:
Addresses issue #645 by hiding the canvas that was causing the issues (since the canvas actually isn't really needed in the iOS app anyways, as it just renders the little white lines pointing to each app icon when you hover your mouse over the recently used tool icon bar).

I've tested that this fixes the issue, but I'm not quite satisfied with this solution since I don't understand why this canvas would be causing issues in the first place, as it doesn't have any blend mode applied to it as far as I can see.

Any ideas as to what the root cause might be? I can look into it further another day, but I'll open this thread for now.

### DIFF
--- a/src/gui/recentlyUsedBar.js
+++ b/src/gui/recentlyUsedBar.js
@@ -362,8 +362,28 @@ class RecentlyUsedBar {
         }
     }
 
+    // we hide the ru-canvas in the AR app, to fix a very strange blend-mode bug noticed on iOS 17.4, iPhone 15 Pro Max
+    fixDeviceSpecificBlendModeBug() {
+        if (realityEditor.device.environment.isWithinToolboxApp()) {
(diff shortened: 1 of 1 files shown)


ptc-rdeleeuw (commented):
I don't have the device to test this / replicate this behavior. For that reason alone I can't Approve the changes. That being said, the code looks good and is properly documented so I won't block this from merging. If @Steve-KX-RL can verify this to be working, I would recommend to merge it.

hobinjk-ptc:
(with the removal of canvasHasContext in a recent PR I believe this is resolved)

benptc:
Oh let me check again with the new main branches

benptc:
Yeah looks like it's fixed now!

PR #647, opened 2024-03-13
Human tracking filtering low confidence joints

ptc-mklaudiny:
General idea is to filter out joints which have a low confidence. The confidence is essentially a visibility score we get from mediapipe for limbs (not torso and head). The filtered joints and adjacent bones are not displayed. Also, some analytics calculations are affected by this (eg. REBA score of a leg should not be calculated if ankle joint is filtered. We then default to the lowest score 1). However, it is not relevant for all lens (Waste/Value does not care about pose directly).

I tried to make the changes not very invasive and shared by all relevant analytics lens. All joints and bones in Pose class has new valid attribute which is calculated based on a given confidence threshold (or recalculated when the threshold is changed on the fly). This attribute is used to show/hide body parts in HumanRenderInstance. Also, it is used in lens-specific calculations such as rebaScore.js.  HPA.setJointConfidenceThreshold(confidence) tries to dynamically propagate new state of joint/bone validity and lens metrics to relevant analytics elements. 
I'm tuning the right level of confidence threshold. Eventually, input text field in analytics settings will be swapped for a checkbox whether to apply filtering with own internal threshold. 

This is to check if you see any conceptual issues with how this is implemented. It influences most parts of analytics tool (skeletons, spaghetti, region cards ...), so let me know if there are still corners where this filtering might not be propagated properly. Missing bits on my list:
- Acceleration lens
- Immediate refresh of pinned region cards when the confidence threshold changes
- Immediate refresh of temporary region card for new time interval when the confidence threshold changes

### DIFF
--- a/src/humanPose/AccelerationLens.js
+++ b/src/humanPose/AccelerationLens.js
@@ -3,8 +3,11 @@ import {MotionStudyLens} from "./MotionStudyLens.js";
 import {MotionStudyColors} from "./MotionStudyColors.js";
 import {JOINTS} from "./constants.js";
 
-const HIGH_CUTOFF = 35 // In m/s^2
-const MED_CUTOFF = 15 // In m/s^2
+const HIGH_CUTOFF = 35; // in m/s^2
+const MED_CUTOFF = 15; // in m/s^2
+
+export const MIN_ACCELERATION = 0; // in m/s^2
+export const MAX_ACCELERATION = 40; // in m/s^2
 
 /**
  * AccelerationLens is a lens that calculates the acceleration of each joint in the pose history.
--- a/src/humanPose/AccelerationLens.js
+++ b/src/humanPose/AccelerationLens.js
@@ -43,26 +43,51 @@ export class AccelerationLens extends MotionStudyLens {
         const previousPreviousPose = previousPose ? previousPose.metadata.previousPose : null;
         if (!previousPose) {
             pose.forEachJoint(joint => {
-                joint.velocity = new THREE.Vector3(); // Velocity is zero for the first pose
+                // Velocity and acceleration are zero for the first pose
+                joint.velocity = new THREE.Vector3(); 
                 joint.speed = 0;
-                joint.acceleration = new THREE.Vector3(); // Acceleration is zero for the first two poses
+                joint.acceleration = new THREE.Vector3();
                 joint.accelerationMagnitude = 0;
             });
         } else if (!previousPreviousPose) {
             pose.forEachJoint(joint => {
                 const previousJoint = previousPose.getJoint(joint.name);
-                joint.velocity = joint.position.clone().sub(previousJoint.position).divideScalar(pose.timestamp - previousPose.timestamp); // mm/ms = m/s
-                joint.speed = joint.velocity.length();
-                joint.acceleration = new THREE.Vector3(); // Acceleration is zero for the first two poses
-                joint.accelerationMagnitude = 0;
+                if (!joint.valid || !previousJoint.valid) {
+                    // Velocity and acceleration are zero if a joint is invalid in current or previous pose
+                    joint.velocity = new THREE.Vector3();
+                    joint.speed = 0;
+                    joint.acceleration = new THREE.Vector3();
+                    joint.accelerationMagnitude = 0;
+                } else {  // joint.valid == true && previousJoint.valid == true 
+                    joint.velocity = joint.position.clone().sub(previousJoint.position).divideScalar(pose.timestamp - previousPose.timestamp); // mm/ms = m/s
+                    joint.speed = joint.velocity.length();
+                    // Acceleration is zero for the second pose
+                    joint.acceleration = new THREE.Vector3(); 
+                    joint.accelerationMagnitude = 0;
+                }
             });
         } else {
             pose.forEachJoint(joint => {
                 const previousJoint = previousPose.getJoint(joint.name);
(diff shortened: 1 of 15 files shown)


src/humanPose/HumanPoseAnalyzer.js
```diff
@@ -1,5 +1,5 @@
 import * as THREE from '../../thirdPartyCode/three/three.module.js';
-import {JOINTS} from './constants.js';
+import {DISPLAY_INVALID_ELEMENTS, JOINTS} from './constants.js';
```
dangond-ptc, line 2: Lint error, should remove this import if it's not being used.

src/humanPose/HumanPoseRenderInstance.js
```diff
@@ -208,10 +211,18 @@ export class HumanPoseRenderInstance {
...
+                this.lensColors[lens.name].joints[JOINT_TO_INDEX[joint.name]] = MotionStudyColors.undefined;
+            }
         });
         this.pose.forEachBone(bone => {
             this.lensColors[lens.name].bones[BONE_TO_INDEX[bone.name]] = lens.getColorForBone(bone);
+            if (!bone.valid) {
+                this.lensColors[lens.name].bones[BONE_TO_INDEX[bone.name]] = MotionStudyColors.undefined;
+            }
         });
+        // MK - why this condition (lens === this.lens)? When switching lens this is not true and this is not applied.
```
dangond-ptc, line 224: It's been a while, so I'm not 100% sure, but I think this was meant for live pose data so that when a new pose is added it only updates the selected lens joint colors for efficiency?

dangond-ptc (approved):
Looks good to me, just check to make sure the lint errors are all caught via `eslint .`

src/humanPose/HumanPoseAnalyzerSettingsUi.js
```diff
@@ -28,6 +28,10 @@ export class HumanPoseAnalyzerSettingsUi {
                             <div class="hpa-settings-section-row-label">View auxiliary poses</div>
                             <input type="checkbox" class="hpa-settings-section-row-checkbox" id="hpa-settings-toggle-child-human-poses">
                         </div>
+                        <div class="hpa-settings-section-row">
+                            <div class="hpa-settings-section-row-label">Set joint confidence</div>
+                            <input type="text" id="hpa-settings-set-joint-confidence">
```
hobinjk-ptc, line 33: input type="number" could be a better choice here (might need to create some css styles to make it fit in)

hobinjk-ptc (commented):
nice docs and organization of the REBA scoring code!

src/humanPose/AccelerationLens.js
```diff
@@ -32,34 +35,59 @@ export class AccelerationLens extends MotionStudyLens {
      * @return {boolean} True if the joint has had its acceleration calculated, false otherwise.
      */
     accelerationAppliedToJoint(joint) {
-        return joint.acceleration && joint.accelerationMagnitude > 0.000001
+        return joint.acceleration && joint.accelerationMagnitude && joint.accelerationMagnitude !== -1
```
hobinjk-ptc, line 38: nit: missing semicolon

src/humanPose/AccelerationLens.js
```diff
@@ -97,23 +125,24 @@ export class AccelerationLens extends MotionStudyLens {
...
+        if (!this.accelerationAppliedToJoint(joint)) {
             return MotionStudyColors.undefined;
         }
         const acceleration = joint.accelerationMagnitude;
         return this.getColorForAcceleration(acceleration);
     }
 
     getColorForBone(bone) {
-        if (typeof bone.joint0.accelerationMagnitude !== "number" || typeof bone.joint1.accelerationMagnitude !== "number") {
+        if (!this.accelerationAppliedToJoint(bone.joint0) || !this.accelerationAppliedToJoint(bone.joint1)) {
```
hobinjk-ptc, line 136: nice simplification! definitely helps performance to avoid polymorphism too!

hobinjk-ptc (approved):

src/humanPose/AccelerationLens.js
```diff
@@ -32,34 +35,59 @@ export class AccelerationLens extends MotionStudyLens {
...
             pose.forEachJoint(joint => {
                 const previousJoint = previousPose.getJoint(joint.name);
-                joint.velocity = joint.position.clone().sub(previousJoint.position).divideScalar(pose.timestamp - previousPose.timestamp); // mm/ms = m/s
-                joint.speed = joint.velocity.length();
-                joint.acceleration = new THREE.Vector3(); // Acceleration is zero for the first two poses
-                joint.accelerationMagnitude = 0;
+                if (!joint.valid || !previousJoint.valid) {
+                    // Velocity and acceleration are zero if a joint is invalid in current or previous pose
+                    joint.velocity = new THREE.Vector3();
+                    joint.speed = -1;
```
ddgond, line 58: Kind of a nit, but I wonder if it would be better to use null instead of -1. ~~Otherwise, someone writing code around this in the future might not realize to check for -1 and do some math which will work because -1 is a number.~~ Actually, js casts null to 0 when doing math with it, so this point is moot. I have a preference for null instead of -1, but this is really just a nit, feel free to ignore.

ddgond (approved):

src/humanPose/AccelerationLens.js
hobinjk-ptc: It's super minor, but any use of null causes deoptimization from the value's slot having to be either a number or null

PR #649, opened 2024-03-15
Refactor main scene renderer

ptc-rdeleeuw:
Refactoring the renderer to isolate the renderer, ground floor, tracked environment and the camera in preparation to implement the changed render loop, global scene scaling and networking.

The main parts are isolated from threejsScene, but some pieces of code have come from different places. the main goal is to eliminate getInternalObject usage as much as possible. This makes sure all logic that makes changes to these objects are concentrated in one class. 
The ground plane and scanned object/anchored group will be used by the network graphics to add and remove objects to the scene. 
The renderer is isolated in preparation for scene scaling and renderloop changes. Both will be managed by this class.
The camera has been isolated in preparation of WebXR which overrides the camera used for rendering with its own camera from the WebXRManager (stereoscopic ArrayCamera for both eyes). Concentrating this logic makes it easier to check for feature parity between both cameras.

The network graphics will add datastorage/synchronization classes to these classes to synchronize data between tool and this renderer. This might make it seem some classes look simple at the moment.

this pull request also contains a fix for a small regression, the VisualDiff code contained a dependency on the  old layered render code, I tried to resolve this without layers but ended up reintroducing layers but skipping layer 1 and 2 for WebXR compatibility. The single layer render still is an improvement over the layered rendering, by reducing scene render calls.


src/gui/scene/GroundPlane.js
```diff
@@ -0,0 +1,63 @@
+import * as THREE from "../../../thirdPartyCode/three/three.module.js"
+
+/**
+ * @typedef {number} Milimeters
```
hobinjk-ptc, line 4: minor typo: Millimeters (two l's)

I like using types like this to make the units clearer!

src/gui/threejsScene.js
```diff
@@ -1234,18 +1117,10 @@ import { MapShaderSettingsUI } from "../measure/mapShaderSettingsUI.js";
     }
 
     /**
-     * @return {{
-           camera: THREE.PerspectiveCamera,
-           renderer: THREE.WebGLRenderer,
-           scene: THREE.Scene,
-       }} Various internal objects necessary for advanced (hacky) functions
+     * @return {mainRenderer} Various internal objects necessary for advanced (hacky) functions
```
hobinjk-ptc, line 1120: The type here should be {Renderer}, the type of the mainRenderer variable

benptc:
I think it's missing one additional `.getInternalObject()` in device/index.js – when I tried to draw, the pointer event posted into the drawing tool calculates the intersect using the groundPlaneCollider

https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/9208f3813d433cfb61e20515a22a457f6784dc06/src/device/index.js#L344

benptc:
Looks good! Just doing some final testing on my end to verify functionality in iOS app + remote operator scenario

benptc (approved):
Code looks good, and everything worked that I was able to test

PR #651, opened 2024-03-18
Splat viewer, refactored avatarIconList, avatar lockOnMode, debugged publicData subscription, support for renderMode, comments and cleanup

benptc:
🆘 I humbly present a very large PR for review 🆘  I'll open the corresponding PRs to remote-operator addon and pop-up addon as soon as I finish wrangling those ones up into a clean enough state.

Goes with the (already merged) https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/1017


src/avatar/iconMenu.js
```diff
@@ -0,0 +1,375 @@
...
+        let sortedKeys = realityEditor.avatar.utils.sortAvatarList(connectedAvatars);
+
+        // if too many collaborators, show a "+N..." at the end (I'm calling this the ellipsis) and limit how many icons
+        const MAX_ICONS = realityEditor.device.environment.variables.maxAvatarIcons;
+
+        // build and add each of the icons to the container, and attach pointer event listeners to them
+        sortedKeys.forEach((objectKey, index) => {
+            let isEllipsis = index === (MAX_ICONS - 1) && sortedKeys.length > MAX_ICONS; // last one turns into "+2", "+3", etc
+            let numTooMany = sortedKeys.length - (MAX_ICONS - 1);
+            if (index >= MAX_ICONS) { return; } // after the ellipsis, we ignore the rest
```
dangond-ptc, line 85: Nit, but this exit would read better at the top of the loop.

src/device/environment.js
```diff
@@ -79,7 +79,7 @@ createNameSpace("realityEditor.device.environment");
         transformControlsSize: 1, // on remote operator, we can scale down the gizmo size for moving groundplane anchors
         localServerPort: 49369, // the port where a local vuforia-spatial-edge-server can be expected
         screenTopOffset: 0, // if there's a menubar on the top, increase this
-        maxAvatarIcons: 5, // limits the number of circular icons depicting how many avatars are currently connected
+        maxAvatarIcons: 7, // limits the number of circular icons depicting how many avatars are currently connected
```
dangond-ptc, line 82: I wonder if it would make sense to change this based on device type? Does 7 look alright on the iPhone?

src/worldObjects.js
```diff
@@ -48,7 +48,7 @@ createNameSpace("realityEditor.worldObjects");
         // when an object is detected, check if we need to add a world object for its server
         realityEditor.network.addObjectDiscoveredCallback(function(object, objectKey) {
             
-            if (object.isWorldObject) {
+            if (object.isWorldObject || object.type === 'world') {
```
dangond-ptc, line 51: Why do we need to check for both of these things? Are we migrating away from using .isWorldObject?

dangond-ptc (approved):
Looks good to me, just a few nits

src/device/environment.js
benptc: Agreed – there's a line in the pop-up-onboarding-addon's index.js that reduces the number to 3 on mobile devices. Eventually we might want something even smarter to maximize usage of the available screen real estate, but I think that'll do for now.

        isRemoteOperator = !realityEditor.device.environment.isARMode();
        if (!isRemoteOperator) {
            ...
            env.maxAvatarIcons = 3;
        }

src/worldObjects.js
benptc: Yeah there used to not be an object `type` property so we added `isWorldObject`, `isAnchor`, etc.. I want to eventually just use `type` for everything as a more generic property (used for avatar, human, world, anchor, object, and whatever else we come out with in the future). There's probably a more intentional way to phase out isWorldObject, but for now I've just been adding the extra conditional whenever I encounter it

PR #656, opened 2024-03-22
Analytics: Simultaneous video recording

hobinjk-ptc:
With the latest ios build and the corresponding core-addon PR this should enable the ability to click "Show Spatial Video" while viewing a highlight animation playing to see the corresponding video at that time. Note that there's no handling of the recording device's orientation, suggestions welcome.

### DIFF
--- a/src/humanPose/Animation.js
+++ b/src/humanPose/Animation.js
@@ -1,0 +1,114 @@
+const AnimationState = {
+    video: 'video',
+    noVideo: 'noVideo',
+};
+
+/**
+ * Animation
+ * Organizes all the information about an HPA's animation in one location
+ * Coordinates propagating animation information to the motionStudy and its
+ * video player if present
+ *
+ * Notably hands off control of video playback to that video player during any
+ * region of the recording that it contains
+ *
+ * Time is in ms unless otherwise specified
+ */
+export class Animation {
+    constructor(humanPoseAnalyzer, motionStudy, startTime, endTime) {
+        this.humanPoseAnalyzer = humanPoseAnalyzer;
+        this.motionStudy = motionStudy;
+
+        this.startTime = startTime;
+        this.endTime = endTime;
+
+        this.cursorTime = this.startTime;
+        this.lastUpdate = -1;
+        this.playing = true;
+
+        this.animationState = AnimationState.noVideo;
+    }
+
+    get videoPlayer() {
+        return this.motionStudy.videoPlayer;
+    }
+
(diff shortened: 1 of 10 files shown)


hobinjk-ptc:
As per the other PR, this now also adds labels to analytics sessions

src/humanPose/Animation.js
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        // Adjusted cursorTime that might be tracking a
+        // wobbly-relative-to-real-time video playback
+        let cursorTimeAdj = this.cursorTime;
+        if (this.isVideoPlayerInControl(cursorTimeAdj) && this.playing) {
+            if (this.animationState !== AnimationState.video) {
+                this.startVideoPlayback(cursorTimeAdj);
+            } else {
+                cursorTimeAdj = this.cursorTimeFromVideo();
```
ptc-mklaudiny, line 92: So if I understand correctly, video player does equivalent of 'this.cursorTime += dt;' and the calculation here is overriden.

ptc-mklaudiny:
It seems to me that the last piece for perfect sync between human poses and video frames is missing. Vuforia timestamp for a pose is stored in lastUpdateDataTS in HumanObject. 
https://github.com/ptcrealitylab/vuforia-spatial-edge-server/blob/4c2a3f3797b304c22ab715d54493743e444bcdb9/server.js#L3604-L3605 
This timestamp should be used for Pose instances when replaying them from recorded data.

hobinjk-ptc:
In src/humanPose/index.js I do swap to lastUpdateDataTS https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/main/src/humanPose/index.js#L256 though it's possible that it's getting overwritten somewhere

src/humanPose/Animation.js
hobinjk-ptc: This Animation does `this.cursorTime += dt`. The VideoPlayer uses its color video element's currentTime property which exactly tracks where the video is.  This code `this.cursorTime` with the more-exact video-derived cursorTimeFromVideo() for pose display. It doesn't write to `this.cursorTime` to guarantee that the animation continues eventually. For example, if the video starts buffering and pauses, then `this.cursorTimeFromVideo()` will stop moving forwards as well. Not writing guarantees that the loop will eventually start from the beginning again

ptc-mklaudiny:
I somehow overlooked that. All good.

PR #660, opened 2024-03-26
Human tracking reba score fixes

ptc-mklaudiny:
Fixes to REBA calculation to adhere better to original definition.
Tuned some thresholds which are not explicitly defined by REBA.
Adjusted coloring of some body parts based on their reba score.


hobinjk-ptc (approved):
Seems reasonable. Good, thorough work

Issue #673, opened 2024-04-02
Drawing tool annotations draw above avatar bar


benptc:

benptc:
Fixed by PR #674

PR #674, opened 2024-04-02
move avatar container z index above the fullscreen tool z index

benptc:
Puts avatar icon container above drawings and fullscreen chat.

Fixes issue https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/issues/673. Fullscreen tools have z index 100, but I put it even higher to give it additional priority.

Before:

Screenshot 2024-04-02 at 3 07 29 PM

After:

Screenshot 2024-04-02 at 2 48 45 PM

Before:

Screenshot 2024-04-02 at 3 07 36 PM

After:

Screenshot 2024-04-02 at 2 48 53 PM

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1827,6 +1827,7 @@ Logic Interface CSS
     background-color: rgba(0,0,0,0.5);
     height: 50px;
     border-radius: 25px;
+    z-index: 1000;
 }
 
 .avatarListIcon {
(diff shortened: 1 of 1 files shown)


hobinjk-ptc (changes requested):
While I'm sure it's not doing anything, it would be more consistent to add a corresponding transform: translate(-50%, 0, 1000px) to match how other rules do it

PR #682, opened 2024-04-03
Switch to spatial.ptc.io

hobinjk-ptc:


dangond-ptc (approved):
I wonder if we could do better by looking for /n/ in `window.location`? Allows compatibility with local cloud-proxy for development purposes.

PR #684, opened 2024-04-04
Touchscreen camera buttons

benptc:
Updates the userinterface to support the ability for the addon to temporarily take control of the pointerevents, preventing the avatar pointer or tool icon interactions when camera control mode has been activated: https://github.com/ptcrealitylab/vuforia-spatial-remote-operator-addon/pull/178

### DIFF
--- a/src/device/index.js
+++ b/src/device/index.js
@@ -641,13 +641,43 @@ realityEditor.device.disablePinchToScale = function() {
 };
 
 /**
+ * This system allows any number of modules to mark with a flag that they're
+ * currently claiming the pointer events for a camera control mode.
+ * @type {Object.<string, boolean>}
+ */
+realityEditor.device.manualCameraControlFlags = {};
+/**
+ * @param {string} flagName - the name of the module/reason that the pointer is claimed by
+ */
+realityEditor.device.setFlagForPointerOccupiedByCamera = function(flagName) {
+    realityEditor.device.manualCameraControlFlags[flagName] = true;
+};
+/**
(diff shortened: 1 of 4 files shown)


src/device/index.js
```diff
@@ -640,14 +640,44 @@ realityEditor.device.disablePinchToScale = function() {
     this.editingState.pinchToScaleDisabled = true;
 };
 
+/**
+ * This system allows any number of modules to mark with a flag that they're
+ * currently claiming the pointer events for a camera control mode.
+ * @type {Object.<string, boolean>}
+ */
+realityEditor.device.manualCameraControlFlags = {};
```
dangond-ptc, line 648: nit: given that every entry in this will be set to true and never false, I think it makes more sense to be a list of names rather than an object. Feel free to ignore though!

dangond-ptc (approved):
LGTM!

Steve-KX-RL (approved):
LGTM!

src/device/index.js
benptc: My thought was that making it an object would inherently prevent duplicates, but I do see what you mean

dangond-ptc: We could also get fancy and use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

benptc: Good idea, updated! I forgot that JS has a Set type, this is a nice place to use it

ktheo-ptc (commented):
LGTM!

PR #687, opened 2024-04-09
Always enable depthTest on CameraVis shader

hobinjk-ptc:
Previously depthTest would be disabled in first person mode which drew correctly in the previous system. Now with the depth-based system replacing layers, depthTest is required to draw above the ground plane visualization.

Adjacent to this, the current state draws first person video above the pose renderers and it would be nice (but not required) to be able to see the pose renderers in first person if you have any free time in search of a use @ptc-rdeleeuw

### DIFF
--- a/src/spatialCapture/Shaders.js
+++ b/src/spatialCapture/Shaders.js
@@ -321,7 +321,7 @@ export function createPointCloudMaterial(texture, textureDepth, shaderMode, bord
         vertexShader,
         fragmentShader,
         // blending: THREE.AdditiveBlending,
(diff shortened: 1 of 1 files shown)


ptc-rdeleeuw (approved):
I don't have time now to look into it, if you can make an issue/ticket for it on the board, then I can get back into this sorting problem later and make the poses draw on top of the video.

hobinjk-ptc:
Filed https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/issues/688 as a follow-up

PR #689, opened 2024-04-10
Remove reference to appendPose

hobinjk-ptc:


benptc (approved):
Didn't get a chance to review this but if you'e tested it we could merge it for now

PR #690, opened 2024-04-11
AI Chatbot

Steve-KX-RL:
1. Have a (somewhat) useable conversation with your Azure AI chatbot, after inputting the credentials.
2. Currently supported "spatial actions" that got recorded and fed into AI include: add/reposition/delete/open/minimize/close a tool. Also able to read communication tool messages & make a summary. Ongoing improvements on user join / leave sessions & name change and reading data from spatial analytics tool.
3. When feeding new & historical messages to AI, avatar usernames and tools go through ai/crc.js to encode into a semantic-agnostic alphanumeric string, so that AI process it as a whole.
4. When getting back AI responses & display messages on screen, go through ai/crc.js to decode the string back to human-readable avatar names & tool names.
5. ai/map.js keep an updated list of all the {id, name, scrambled id (crc.js encoded id)} as a reference.
6. Also need edge-server and remote-operator to work.


css/ai.css
```diff
@@ -0,0 +1,142 @@
+body {
+    margin: 0;
```
hobinjk-ptc, line 2: ` margin: 0` is already set on the body here: https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/59137ea3a9fd546dd3074e018806c04f9c033824/css/index.css#L75

css/ai.css
```diff
@@ -0,0 +1,142 @@
...
+    z-index: 1100;
+}
+
+@keyframes slideToLeft {
+    from {
+        transform: translateX(0);
+    }
+    /*to { transform: translateX( calc(-(max(100px, min(400px, 100%)) - 20px - 20px - 20px - 10px)) ); }*/
+    to {
+        transform: translateX( calc(-110%) );
```
hobinjk-ptc, line 36: can drop the unnecessary `calc` here and below

src/ai/index.js
```diff
@@ -0,0 +1,550 @@
...
+                realityEditor.network.postAiApiKeys(endpointArea.value, apiKeyArea.value, true);
+            }
+        })
+        
+        searchTextArea.addEventListener('input', function() {
+            // adjustTextAreaSize();
+        });
+        
+        // todo Steve: below 4 doesn't work
+        searchTextArea.addEventListener('mousedown', (e) => {e.stopPropagation();});
```
hobinjk-ptc, line 383: I believe the reason these aren't working is that they should be pointerdown pointerup pointermove etc.

src/ai/index.js
```diff
@@ -0,0 +1,550 @@
...
+            // realityEditor.device.keyboardEvents.openKeyboard();
+        });
+
+        searchTextArea.addEventListener('blur', () => {
+            // console.log('blur');
+            // realityEditor.device.keyboardEvents.closeKeyboard();
+        });
+
+        searchTextArea.addEventListener('keydown', (e) => {
+            // todo Steve: bug: press g/G still focuses the camera to target. Is it b/c remote operator event listener is added later than searchTextBar listener? Figure out how the orders & layers work
```
hobinjk-ptc, line 407: If you want to test whether it's the order, you can pass `true` as the last argument to the key* event listeners to capture the event earlier in the phase. However, the actual bug causing this problem is that the g/G listener is using 'keypress' and here you look for keydown and keyup

src/humanPose/HumanPoseAnalyzer.js
```diff
@@ -1096,6 +1096,14 @@ export class HumanPoseAnalyzer {
             clone.renderer.markMatrixNeedsUpdate();
         });
     }
+    
+    getDataByTimestamp(timestamp) {
```
hobinjk-ptc, line 1100: This feels mostly unnecessary and could be replaced by getClonesByTimestamp without losing much. `undefined` vs `[]` isn't a big difference.

src/ai/index.js
Steve-KX-RL: Make sense!

src/ai/index.js
Steve-KX-RL: Right. I'll add a proper focus event in the remote operator to make sure it's nicely integrated in the app, instead of making a standalone keypress listener.

hobinjk-ptc (approved):

hobinjk-ptc:
Nice, comprehensive work!

src/network/index.js
```diff
@@ -1008,6 +1027,15 @@ realityEditor.network.onAction = function (action) {
...
         }
+        
+        // todo Steve: repositioned a frame
+        let params = {
+            objectKey: thisFrame.objectId,
+            frameKey: thisFrame.uuid,
+            frameType: thisFrame.src,
+        };
+        let avatarId = realityEditor.avatar.getAvatarObjectKeyFromSessionId(thisAction.lastEditor);
+        realityEditor.ai.onFrameRepositioned(params, avatarId);
```
benptc, line 1038: Similar to above comment, it might be better to subscribe to an event (though looks like one doesn't exist for this yet). However, I'd also recommend you move this code block into the 

```
if (thisKey === 'ar' &&
    propertiesToIgnore.indexOf('ar.x') > -1 &&
    propertiesToIgnore.indexOf('ar.y') > -1) {
    ...
```
part of realityEditor.network.reloadFrame, as where you have it now might get triggered by various updates within the frame, not just repositioning.

benptc (changes requested):
Great work overall, lots of awesome functionality in here! I did leave some comments inline with some suggested small documentation or code changes. I also tested the system and it seems to work for me 💯

css/index.css
```diff
@@ -2283,6 +2283,16 @@ Logic Interface CSS
     cursor: pointer;
 }
 
+.ru-focus {
```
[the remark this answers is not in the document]
Steve-KX-RL: I think this code is originally for the recently used bar. During the last bug fix, we probably no longer use it.

png/focus.png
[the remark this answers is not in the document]
Steve-KX-RL: It was from a past experiment --- I'll delete it.

src/ai/index.js
```diff
@@ -0,0 +1,533 @@
...
+        inverseGroundPlaneMatrix.invert();
+        position.applyMatrix4(inverseGroundPlaneMatrix);
+
+        let timestamp = getFormattedTime();
+        let avatarName = realityEditor.avatar.getAvatarNameFromObjectKey(avatarId);
+        let avatarScrambledId = realityEditor.ai.crc.generateChecksum(avatarId);
+        let frameScrambledId = realityEditor.ai.crc.generateChecksum(frameId);
+        map.addToMap(avatarId, avatarName, avatarScrambledId);
+        map.addToMap(frameId, frameType, frameScrambledId);
+        let newInfo = `User ${avatarScrambledId} added a ${frameScrambledId} tool at ${timestamp} at (${position.x.toFixed(0)},${position.y.toFixed(0)},${position.z.toFixed(0)})`;
```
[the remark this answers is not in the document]
Steve-KX-RL: Yes --- I do plan to leave this information in for future reference. But another way is to directly query object.json and get the coords from there. Probably a future route that I'm going to take, since the current setup doesn't have memory after refreshing the page. Probably need to add more info to object.json tho, eg: who edited the tool, etc.

src/ai/index.js
```diff
@@ -0,0 +1,533 @@
...
+        // todo Steve: after switching from getavatarIdFromSessionId() to getAvatarObjectKeyFromSessionId(), this still stays the old way. Need to change later
+        let _timestamp = getFormattedTime();
+        if (oldName === null) {
+            // let newInfo = `User ${newName} joined the space at ${timestamp}`;
+            // question1 += `\n${newInfo}`;
+        } else {
+            // let newInfo = `User ${oldName} has changed their name to ${newName}`;
+            // question1 += `\n${newInfo}`;
+        }
+    }
```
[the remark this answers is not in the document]
Steve-KX-RL: Yes

src/ai/index.js
```diff
@@ -0,0 +1,533 @@
...
+    let aiContainer;
+    let endpointArea, apiKeyArea;
+    let searchTextArea;
+    let dialogueContainer;
+    
+    let keyPressed = {
+        'Shift': false,
+        'Enter': false,
+    };
+    let INCLUDE_TEST_HISTORY_MESSAGE = true;
```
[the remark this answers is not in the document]
Steve-KX-RL: Oh that was for testing. Will get rid of it.

src/ai/index.js
```diff
@@ -0,0 +1,533 @@
...
+
+        searchTextArea.addEventListener('keyup', (e) => {
+            e.stopPropagation();
+            
+            if (e.key === 'Enter') {
+                keyPressed['Enter'] = false;
+            } else if (e.key === 'Shift') {
+                keyPressed['Shift'] = false;
+            }
+        });
```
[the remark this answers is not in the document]
Steve-KX-RL: 👀Nice catch, thx Ben!

src/network/index.js
```diff
@@ -697,6 +698,13 @@ realityEditor.network.updateObject = function (origin, remote, objectKey) {
             origin.frames[frameKey].uuid = frameKey;
             
             realityEditor.network.initializeDownloadedFrame(objectKey, frameKey, origin.frames[frameKey]);
+            // todo Steve: added a new frame
+            let params = {
+                objectKey: objectKey,
+                frameKey: frameKey,
+                frameType: origin.frames[frameKey].src,
+            };
+            realityEditor.ai.onFrameAdded(params, avatarName);
```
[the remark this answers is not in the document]
Steve-KX-RL: Good idea. I'll update all the event listeners to follow this guideline.

src/network/index.js
Steve-KX-RL: The internal reposition mechanism changed and I can't get the original & updated frame positions from here. Will temporarily disable the "frameRepositioned" event.

src/ai/crc.js
```diff
@@ -1,5 +1,12 @@
 createNameSpace("realityEditor.ai.crc");
 
+/**
+ * @fileOverview realityEditor.ai.crc
+ * Contains a method to convert a tool / avatar id (eg: _WORLD_sessionDPjwiU0r_P4jnv2zz9aaspatialDraw1rXagcg6zvdk0)
+ * into a 6-digit alphanumeric "scrambled id" (eg: 2yM7oX). 
+ * The scrambled id is semantic-agnostic, so that AI process it as a whole unique id.
+ */
```
benptc, line 8: Thanks for the updated documentation!

src/ai/index.js
```diff
@@ -4,27 +4,47 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
...
     
     function focusOnFrame(frameKey) {
         let framePosition = realityEditor.gui.threejsScene.getToolPosition(frameKey);
         let cameraPosition = realityEditor.gui.threejsScene.getCameraPosition();
         let frameDirection = cameraPosition.clone().sub(framePosition).normalize();
-        realityEditor.device.desktopCamera.focusVirtualCamera(framePosition, frameDirection);
+        callbackHandler.triggerCallbacks('shouldFocusVirtualCamera', {
+            pos: {x: framePosition.x, y: framePosition.y, z: framePosition.z},
+            dir: {x: frameDirection.x, y: frameDirection.y, z: frameDirection.z}
+        });
```
benptc, line 47: Nice, thanks for adjusting the implementation this way!

benptc (approved):
Thanks for making all the adjustments / cleaning things up! LGTM now 👍

PR #695, opened 2024-04-16
Global scale & temp vr button

ptc-rdeleeuw:
this pull request is the last hurdle to webxr support (on vision pro)
it scales the environment when between milimeters in the browser and ios to meters in webxr
it also enables webxr in three.js and adds a button to enter vr immersive mode

depends on:
https://github.com/ptcrealitylab/vuforia-spatial-remote-operator-addon/pull/189


src/gui/ar/draw.js
ptc-rdeleeuw, line 1: This line escapes early in the renderloop and therefor prevents rendering when in the immersive mode

src/gui/scene/Renderer.js
```diff
@@ -1,15 +1,103 @@
...
+    #listeners;
+    
+    /**
+     * 
+     * @param {DeviceUnitsPerMeter} deviceScale 
+     * @param {MetersPerSceneUnit} sceneScale 
+     */
+    constructor(deviceScale, sceneScale) {
+        this.#deviceScale = deviceScale;
+        this.#sceneScale = sceneScale;
```
ptc-rdeleeuw, line 40: the scene scale doesn't actually change, but a lot of code depends on the separation of device scale and scene scale, this could be a constant. But should it be, maybe in the remote operator we want to be able to scale environments if we want to display multiple environments at once (minimap idea of the hackathon). Most of the code would already be compatible.

src/gui/scene/Renderer.js
```diff
@@ -25,19 +113,31 @@ class Renderer {
...
+    /** @type {GlobalScale} */
+    #globalScale
+
     /**
      * 
      * @param {HTMLCanvasElement} domElement 
      */
     constructor(domElement) {
-        this.#renderer = new THREE.WebGLRenderer({canvas: domElement, alpha: true, antialias: false});
+        this.#renderer = new THREE.WebGLRenderer({canvas: domElement, alpha: true, antialias: true});
```
ptc-rdeleeuw, line 125: Antialiasing should not have a big impact on the code, if needed we can disable this setting for non webXR environments but for sake of simplicity, I turned it on for all versions.

src/gui/scene/Renderer.js
```diff
@@ -81,14 +181,33 @@ class Renderer {
...
+
+    webXRAvailable() {
+        return this.#renderer.xr && this.#renderer.xr.enabled === true;
+    }
+
+    /**
+     * 
+     * @param {Camera} camera 
+     */
+    setCamera(camera) {
```
ptc-rdeleeuw, line 207: we explicitly set a camera here, so we can choose the camera outside of the renderer. In the future there might be other cameras depending on the hardware platform we're working on. By seperating camera and renderer we can add those more simply. This does get a bit weird in this case, since the webxr camera is managed by the three.js renderer's xr manager. So even if we don't switch the cameras three.js just overrides our choice of camera internally.
The proposed setup will enable us to create an arview camera for the native ios app, a camera with controls for the remote desktop variant and a camera for webxr. At this moment however the first two cameras use the same default class with different configurations.

src/gui/threejsScene.js
```diff
@@ -89,8 +94,10 @@ import {setMatrixFromArray} from "./scene/utils.js";
         const domElement = document.getElementById('mainThreejsCanvas');
         mainRenderer = new Renderer(domElement);
 
-        mainCamera = new DefaultCamera("mainCamera", window.innerWidth / window.innerHeight);
-        mainRenderer.add(mainCamera); // Normally not needed, but needed in order to add child objects relative to camera
+        defaultCamera = new DefaultCamera("Default Camera", window.innerWidth / window.innerHeight);
+        webXRCamera = null; // can only be initilized if we have a webxr session
+        mainRenderer.add(defaultCamera); // Normally not needed, but needed in order to add child objects relative to camera
```
ptc-rdeleeuw, line 99: I don;t know if we should also add the internally managed webxr camera for the same reason, but attaching things to the webxr camera might be a bad idea on its own

src/gui/scene/Renderer.js
```diff
@@ -161,9 +280,12 @@ class Renderer {
         objectsToCheck.forEach(obj => {
             this.#raycaster.layers.mask = this.#raycaster.layers.mask | obj.layers.mask;
         });
-        let results = this.#raycaster.intersectObjects( objectsToCheck || this.#scene.children, true );
+        let results = this.#raycaster.intersectObjects( objectsToCheck || this.#globalScale.getNode().children, true );
         results.forEach(intersection => {
             intersection.rayDirection = this.#raycaster.ray.direction;
+            intersection.scenePoint = intersection.point.clone();
```
ptc-rdeleeuw, line 286: important change to communicate to the team, if you ant to make sure your raycast works in milimeter use theses values (scenePoint and sceneDistance)

hobinjk-ptc: Do these need changes as well then?
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/main/src/device/index.js#L355
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/main/src/device/index.js#L373
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/main/src/avatar/index.js#L461
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/main/src/avatar/index.js#L469
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/main/src/gui/spatialIndicator.js#L175
https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/blob/main/src/gui/ar/groundPlaneRenderer.js#L194

hobinjk-ptc (approved):
Overall LGTM! Local testing on browser and mobile both seem to work as well. One minor thing I'd like to change is to hide the "Enter VR" button by default then we can find a better way to do that flow in the future. As it is, having the button just be there gets in the way of other UI.

My current idea for this would be placing an option to "Enter VR" under the Develop menu of the toolbar

src/gui/scene/Renderer.js
hobinjk-ptc: Great, thanks!

hobinjk-ptc:
Another one would be to add an "Enter VR" button to the right-hand kebab menu

benptc:
Yeah, I'd echo what James said and hide the button by default (particularly if it has the "WebXR Needs HTTPS" message).

Screenshot 2024-04-19 at 1 25 18 PM

I guess my chrome browser has the WebXR device emulator extension, causing the Enter VR button to show, but it would be nice to only show it if the user needs the button, as in the current placement it can overlap with some of the other UI elements like the analytics timeline.

Screenshot 2024-04-19 at 1 26 31 PM

Screenshot 2024-04-19 at 1 28 12 PM

Screenshot 2024-04-19 at 1 28 38 PM (2)

src/gui/scene/Renderer.js
benptc: @Steve-KX-RL and @dangond-ptc  tagging you here for visibility – be aware of the new `scenePoint` and `sceneDistance` that Rik added to the raycasting results – you may need to access the point like `raycastIntersects[0].scenePoint` rather than `raycastIntersects[0].point` in future when working in millimeter scale.

thirdPartyCode/THREE.InfiniteGridHelper/InfiniteGridHelper.module.js
```diff
@@ -121,6 +122,12 @@ class InfiniteGridHelper extends Mesh {
 
     }
 
+    setSizesAndDistance(size1, size2, distance) {
```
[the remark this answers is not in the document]
benptc: I'm ok either way, if you want to add it or remove it.

benptc (approved):
Code LGTM

thirdPartyCode/THREE.InfiniteGridHelper/InfiniteGridHelper.module.js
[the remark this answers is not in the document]
ptc-rdeleeuw: I'll keep it

src/gui/scene/Renderer.js
```diff
@@ -1,15 +1,103 @@
...
+    #deviceScale;
+
+    /** @type {MetersPerSceneUnit} */
+    #sceneScale;
+
+    /** @type {THREE.Group} */
+    #node;
+
+    /** @type {GlobalScaleListener[]} */
+    #listeners;
```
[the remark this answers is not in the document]
ptc-rdeleeuw: I'll keep it

ptc-rdeleeuw:
this pull request now depends on https://github.com/ptcrealitylab/vuforia-spatial-remote-operator-addon/pull/189 to make sure the menu item button works

hobinjk-ptc (approved):
Great! Having the button in the menu is an excellent improvement

PR #696, opened 2024-04-16
Update AI module to use cloud-proxy-based language service

dangond-ptc:
Will only work if connected to cloud-proxy, and cloud-proxy has been updated

### DIFF
--- a/src/ai/index.js
+++ b/src/ai/index.js
@@ -239,6 +239,7 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         adjustTextAreaSize();
         setupEventListeners();
         setupSystemEventListeners();
+        hideEndpointApiKeyAndShowSearchTextArea();
     }
     
     function authorSetToString(authorSet) {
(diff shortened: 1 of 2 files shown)


src/ai/index.js
```diff
@@ -239,6 +239,7 @@ import * as THREE from '../../thirdPartyCode/three/three.module.js';
         adjustTextAreaSize();
         setupEventListeners();
         setupSystemEventListeners();
+        hideEndpointApiKeyAndShowSearchTextArea();
```
dangond-ptc, line 242: Calling this immediately so that it skips the ApiKey step. Should probably just remove that code entirely.

src/network/index.js
```diff
@@ -2413,29 +2413,26 @@ realityEditor.network.postAiApiKeys = function(endpoint, azureApiKey, isInit = f
...
-    let ip = worldId.ip;
-    let port = realityEditor.network.getPort(worldId);
+realityEditor.network.postQuestionToAI = async function(conversation, extra) {
     let route = '/ai/question';
     
-    this.postData(realityEditor.network.getURL(ip, port, route), 
-        {
+    const response = await fetch(window.location.origin + route, {
+        headers: {
+            'Accept': 'application/json',
```
hobinjk-ptc, line 2421: This will work with the code I pointed out in cloud-proxy if you add
Authorization: `Bearer ${window.localStorage.something}`

PR #697, opened 2024-04-16
added creation of notification container and adjusted style

ktheo-ptc:
- restored creation of notification container
- reduced width of top screen notification to 50% to avoid conflicts with other UI


benptc (approved):
Tested, working for me!

PR #706, opened 2024-04-26
fixes bug where analytics tool doesnt delete properly in some cases

benptc:
Bug happens if you add it, minimize it without recording anything, and drag the icon to the trash. The video player is null in that case.


hobinjk-ptc (approved):
Thanks for fixing this!

PR #707, opened 2024-04-29
Hide videoPlayer's camera mesh along with its point cloud

hobinjk-ptc:


ptc-mklaudiny (approved):
LGTM. Thanks for the fix

PR #708, opened 2024-04-29
GS Raycasting

Steve-KX-RL:
Implemented raycasting on Gaussian Splatting models, with some tricks to keep the FPS almost the same as the non-raycasting version while navigating in the 3D scene. Need remote-operator to work.
1. used an off-screen FrameBufferObject to perform accurate mouse raycasting in the user-interface splatting.js and spatialCursor/index.js, and pass the information down to tool iframes
2. FrameBufferObject texture resizes when window resizes, to support accurate mouse raycasting
3. During low-FPS (< 30), rolls back to traditional raycasting
4. toggle on/off raycasting in remote-operator/VirtualCamera.js
5. Fixed 2 small typos in device/index.js, when passing raycast info down to tool iframes, pass the original raycast points instead of scaled scenePoints (raycastIntersects[0].point.clone().applyMatrix4(inverseGroundPlaneMatrix);


src/splatting/Splatting.js
```diff
@@ -599,27 +617,85 @@ void main () {
...
+    vec4 actualCam = cam;
+    vec4 offset1 = vec4(majorAxisOffset * (abs(position.x) * majorAxis / viewport), 0., 0.);
+    offset1 = inverse(projection) * (offset1 * pos2d.w);
+    vec4 offset2 = vec4(minorAxisOffset * (abs(position.y) * minorAxis / viewport), 0., 0.);
+    offset2 = inverse(projection) * (offset2 * pos2d.w);
+    actualCam = actualCam + offset1 + offset2;
+    vFBOPosColor = inverse(view) * actualCam;
+    vFBOPosColor.a = 1.;
+    
+    if (vFBOPosColor.xyz == uCollidePosition) { // todo Steve new 11: directly compare the position here and see if it works
```
ptc-rdeleeuw, line 634: Has an issue been created for this todo? Otherwise this will be forgotten and if it works it might be better to remove the comment.

ptc-rdeleeuw:
Didn't test but the raycasting seem ok to me

ptc-rdeleeuw (approved):
Didn't test, but the raycasting code seems logical

src/splatting/Splatting.js
Steve-KX-RL: Oh I forgot to delete it

benptc (approved):
Performance is very good on my machine! (though I do have a very powerful machine so it's probably not the best way to measure).

I feel like the drawings can end up very wiggly, so we might want to experiment with the drawing tool to figure out if theres a better way to smooth things when in GS mode.

I didn't have enough time to thoroughly review the code, but overall it looks good, so I think it's ok to merge if everyone else approves, too

src/splatting/index.js
```diff
@@ -0,0 +1,11 @@
...
+
+createNameSpace("realityEditor.splatting");
+
+(function(exports) {
+    
+    exports.toggleGSRaycast = SplatRenderer.toggleGSRaycast;
+    exports.showGSSettingsPanel = SplatRenderer.showGSSettingsPanel;
+    exports.hideGSSettingsPanel = SplatRenderer.hideGSSettingsPanel;
+    
+}(realityEditor.splatting));
```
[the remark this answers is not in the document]
Steve-KX-RL: Yea looks like this is a more convenient option

PR #711, opened 2024-04-30
Motion study better hydration

hobinjk-ptc:
Fixes an issue where the first load of a motion study after opening a page would feature regionCards without any pose data

### DIFF
--- a/src/motionStudy/motionStudy.js
+++ b/src/motionStudy/motionStudy.js
@@ -531,6 +531,8 @@ export class MotionStudy {
             regionCard.removePinAnimation();
             this.addRegionCard(regionCard);
         }
+
+        this.sortPinnedRegionCards();
     }
 
     createVideoPlayerShowHideButton() {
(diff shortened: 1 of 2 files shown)


src/motionStudy/motionStudy.js
```diff
@@ -564,14 +568,21 @@ export class MotionStudy {
         for (let pinnedRegionCard of this.pinnedRegionCards) {
             if ((Math.abs(pinnedRegionCard.startTime - regionCard.startTime) < tolerance) &&
                (Math.abs(pinnedRegionCard.endTime - regionCard.endTime) < tolerance)) {
-                // New region card already exists in the list
-                regionCard.remove();
+                // New region card already exists in the list, remove it but
```
ptc-mklaudiny, line 571: Is this check of the 'same' card based on nearly similar start/end timestamps still necessary? Timestamps are taken from poses and they should not change when poses are shifted from live to historical pose list. 
When do we get into this if block? Is it when someone stops recording poses and live marking steps on mobile device?

ptc-mklaudiny (approved):
Seems ok. I just put in some question for understanding.

src/motionStudy/motionStudy.js
hobinjk-ptc: You're correct, it probably isn't necessary. This code is a lot more fragile than I'd like since it's trying to deal with some annoying concurrent editing cases. The main use it fulfills is whenever an edit happens on one client (e.g. live marking a step or adjusting a region card after recording) every other client gets a message containing all the region cards which results in this consolidation function. Notably it prefers the existing region card for cases where User A is editing a region card's name and User B is also editing a different region card's name and recreating the region card would interrupt the cursor position.

I'm going to take some time to think about this more since I think it's gotten a bit too convoluted and has some other bugs I've noticed while really diving into it

ptc-mklaudiny (approved):
LGTM.

PR #714, opened 2024-05-03
Add part of solution for portrait videos; Flip iPad video to match testflight orientation

hobinjk-ptc:
Still need to think of a way to determine if the raw, oriented-relative-to-vuforia's-idea-of-the-world video came from portrait or landscape recording but this at least flips iPad video while laying the groundwork

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -2049,7 +2049,7 @@ Logic Interface CSS
     bottom: calc(320px + 91px); /* guess at video top */
 }
 
-.analytics-video {
+.analytics-video, .analytics-video-background {
     position: absolute;
     left: 35px;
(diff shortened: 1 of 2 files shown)


ptc-mklaudiny (commented):
If you mean the controls directly inside the video inset, I mostly used them for checking anonymisation (I maximise the video and use those controls). At the moment, I don't have portrait videos, because of the recording issue on iPhone.

The orientation issue is tricky, especially when we enable both portrait and landscape on iPad. Then a user can change orientation even during recording a single video. 
A way of going around it could be a manual rotation of video by a user during playback. A button for 90deg clockwise which can step to all 4 orientations.

ptc-mklaudiny (approved):
LGTM

PR #719, opened 2024-05-10
Human analytics extra data from reba

ptc-mklaudiny:
This PR refactors REBA calculations to separate generic ergonomic data (angles, 3d offsets) which can be used to calculate other scores like MURI. In principle, this already contains all angles needed by MURI.
REBA calculation has simpler logic which does score counting based on ergonomic data above. 
REBA is more easily configurable in terms of strain levels and their visualisation.
This is a groundwork for introducing MURI implementation in the next step. 

One question how to pass the base ergonomics angles towards export. We store reba scores/colors in all joints/bones of Pose object which has a fair amount of duplication. But it kind of makes sense that a score of the body part is assigned to its all joints/bones.  With these angles it makes sense less and I think it would be good to store them in just one place in Pose. This could be:
1. at the relevant joint (not always clear cut)
```
"joints": {
      "left_elbow":{
           "rebaScore":1,
           "left_lower_arm_bend_angle": 30.0   
```  
2. all angles together outside of 'joints' object    
```
"metadata": {
      "left_lower_arm_bend_angle": 30.0,
     "right_lower_arm_bend_angle": 30.0,
}  
```
[UPDATE] I went with the option 2.


hobinjk-ptc:
Yeah, I'm a fan of the second option. Especially as criteria become derived from more and more joints I think this just makes sense

src/humanPose/ErgonomicsData.js
```diff
@@ -0,0 +1,263 @@
...
+    calculate() {
+        this.calculateOrientations();
+        this.calculateAngles();
+        this.calculateOffsets();
+    }
+    
+    calculateOrientations() {
+        // make sure all coord systems have orthogonal and unit axes
+        // hips are considered a root of the whole body
+        this.orientations.hips.up.set(0, 1, 0); // Hips do not really have an up direction (i.e., even when sitting, the hips are always up), hence given a vector opposite to gravity
```
hobinjk-ptc, line 144: an important philosophical question, answered

hobinjk-ptc (approved):
Overall LGTM. Great reorganizations to have!

hobinjk-ptc (approved):

dangond-ptc (approved):
Very nice changes!

PR #722, opened 2024-05-26
Human analytics muri score

ptc-mklaudiny:
Contains calculation of Muri score as defined by Volvo.
Adds this ergonomic score to analytic lenses.
Calculates evolved version of statistics included in current Muri Excel sheet used in Volvo. These expand a lot the current CSV export. They can be repurposed for other types of future exports or visualisations within Toolbox. 
Cleaned up few things in REBA score which came up during implementation of Muri score.

### DIFF
--- a/src/humanPose/MuriLens.js
+++ b/src/humanPose/MuriLens.js
@@ -1,7 +1,7 @@
 import {MotionStudyLens} from "./MotionStudyLens.js";
 import {MuriScore} from "./MuriScore.js";
 import {MotionStudyColors} from "./MotionStudyColors.js";
-import {JOINTS, ERGO_ANGLES, ERGO_OFFSETS} from "./constants.js";
+import {ERGO_ANGLES, ERGO_OFFSETS} from "./constants.js";
 
 /**
  * MuriLens is a lens that calculates the Muri score for each pose in the history. Individual components of the score
--- a/src/humanPose/MuriLens.js
+++ b/src/humanPose/MuriLens.js
@@ -42,7 +42,7 @@ export class MuriLens extends MotionStudyLens {
                 pose.metadata.ergonomics[angleName] = null;
             }
         });
-        // TODO?: Volvo has bit different definition of this angle (straight arm has 180 deg). However, this definition is inconsistent with other angles which are zero in the netural pose. 
+        // TODO?: Volvo has bit different definition of this angle (straight arm has 180 deg). However, this definition is not properly established and is inconsistent with other angles which are zero in the netural pose. 
         // pose.metadata[ERGO_ANGLES.LEFT_LOWER_ARM_BEND] = 180 - pose.metadata[ERGO_ANGLES.LEFT_LOWER_ARM_BEND];
         // pose.metadata[ERGO_ANGLES.RIGHT_LOWER_ARM_BEND] = 180 - pose.metadata[ERGO_ANGLES.RIGHT_LOWER_ARM_BEND];
         Object.values(ERGO_OFFSETS).forEach(offsetName => {
--- a/src/humanPose/MuriLens.js
+++ b/src/humanPose/MuriLens.js
@@ -18,8 +18,7 @@ export class MuriLens extends MotionStudyLens {
     }
     
     applyLensToPose(pose, force = false) {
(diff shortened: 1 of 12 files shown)


src/humanPose/MuriLens.js
```diff
@@ -0,0 +1,101 @@
...
+        Object.values(ERGO_ANGLES).forEach(angleName => {
+            // angles which could not be calculated are explicitly stored as null
+            if (ergonomicsData.angles.hasOwnProperty(angleName)) {
+                pose.metadata.ergonomics[angleName] = ergonomicsData.angles[angleName];
+            }
+            else {
+                pose.metadata.ergonomics[angleName] = null;
+            }
+        });
+        // TODO?: Volvo has bit different definition of this angle (straight arm has 180 deg). However, this definition is not properly established and is inconsistent with other angles which are zero in the netural pose.
```
dangond-ptc, line 45: What would you recommend for this moving forward? Should we have some mechanism for Volvo to use their definition while keeping it as 0-based for everyone else?

src/humanPose/MuriScore.js
```diff
@@ -0,0 +1,818 @@
...
+    upperArmFrontRaiseAngleThresholds: [60, 90], // degrees
+    upperArmSideRaiseAngleThresholds: [60, 90], // degrees
+    lowerArmBendAngleThresholds: [120], // degrees
+    lowerArmTwistAngleThresholds: [120],  // degrees
+    handFrontBendAngleThresholds: [-20, 30], // degrees
+    handReachDistanceThresholds: [300, 500],  // mm   //TODO: tune based on customer feedback
+
+    // Weigths for different levels of strain on a body part which are used to calculate actual score.
+    // Note: Weights are integer numbers in the latest iteration of MURI method
+    scoreWeights: [0, 1, 2, 5],
```
dangond-ptc, line 24: Where do these come from?

src/humanPose/MuriScore.js
```diff
@@ -0,0 +1,818 @@
...
+
+/** Modifiable configuration of thresholds for MURI calculation. */
+export const MURI_CONFIG = Object.assign({}, MURI_CONFIG_DEFAULT);
+
+const MURI_COLOR_START = MotionStudyColors.fade(MotionStudyColors.green);
+const MURI_COLOR_END = MotionStudyColors.fade(MotionStudyColors.red);
+
+export const MIN_MURI_SCORE = 0
+// Note: calculate precisely based on final calculation of overall muri score.
+export const MAX_MURI_SCORE = 6 * MURI_CONFIG_DEFAULT.scoreWeights[2] /* trunk + head */ + 2 * 6 * MURI_CONFIG_DEFAULT.scoreWeights[2] /* arms */ + 1 * MURI_CONFIG_DEFAULT.scoreWeights[3] /* legs */ // == 41
```
dangond-ptc, line 49: Should the first one here be `scoreWeights[1]`?

dangond-ptc (approved):
Looks good to me! Just left a few questions for my own understanding of the Muri approach.

src/humanPose/MuriScore.js
hobinjk-ptc: I think this could be better explained as "MAX_MURI_SCORE_TRUNK_HEAD = 6 joints * max score for those 6 joints is scoreWeights[2]" and so on and so forth. Refactoring out to individual variables for each part might help, also explaining the source of each magic number

src/humanPose/MuriLens.js
ptc-mklaudiny: For Volvo, this is a new strain type which they introduced recently, so it's definition is not well established. Also, I think they are flexible to some extent to adjust their Muri analysis. So my recommendation is to suggest them to adopt our definition.
Moreover, Valentin wants to keep most of the data within Toolbox based on the current direction. So these raw angle values may not be exposed at all. So as long as Volvo sees the right scores/colors for their angle ranges in our UI, it does not matter what exactly is the underlying value (angle or (180 - angle) )

src/humanPose/MuriScore.js
ptc-mklaudiny: These are score weights defined in Volvo's MURI Excel sheet for different levels of severity.

PR #732, opened 2024-06-24
Add raycasting against tools

dangond-ptc:
Requires https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/1091

### DIFF
--- a/src/device/index.js
+++ b/src/device/index.js
@@ -135,6 +135,9 @@ realityEditor.device.previousPointerMove = null;
  */
 realityEditor.device.callbackHandler = new realityEditor.moduleCallbacks.CallbackHandler('device/index');
 
+realityEditor.device.raycastIndex = 0; // Used to ensure raycast requests don't override each other
+realityEditor.device.raycastCallbacks = new Map();
+
 /**
  * Adds a callback function that will be invoked when the specified function is called
  * @param {string} functionName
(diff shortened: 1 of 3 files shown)


src/device/index.js
```diff
@@ -392,6 +406,74 @@ realityEditor.device.postEventIntoIframe = function(event, frameKey, nodeKey) {
...
+            let resolved = false;
+
+            const key = object + frame + currentIndex;
+            setTimeout(() => {
+                realityEditor.device.raycastCallbacks.delete(key);
+                if (!resolved) {
+                    resolved = true;
+                    reject();
+                }
+            }, 200); // Reject automatically within 200 ms
```
Steve-KX-RL, line 445: https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/assets/111802681/3759d415-6593-47de-a46f-07b8907dfc66

Looks like when I draw very fast, it raycasts on the table first and then flies off. When I draw slower, it forms these squiggly lines. Is it due to this 200ms interval where we decide to listen for tool raycasts?

dangond-ptc: Close, that's a side-effect of how the drawing tool draws that I overlooked when adding the caching mechanism on the object.js side in the edge-server. Removing the caching doesn't seem to cause any performance issues, so I just removed that instead. This is just here to ensure a badly-behaved tool doesn't cause tons of touch events to be delayed.

Steve-KX-RL: Got it. I pulled that commit and no longer get this artifact.

Steve-KX-RL (approved):
LGTM!

src/device/index.js
```diff
@@ -392,6 +406,74 @@ realityEditor.device.postEventIntoIframe = function(event, frameKey, nodeKey) {
...
+
+            realityEditor.device.raycastCallbacks.set(key, (result) => {
+                if (!resolved) {
+                    resolved = true;
+                    resolve(result);
+                }
+            });
+        })
+    });
+    const raycastResults = await Promise.all(frameRaycastPromises);
```
Steve-KX-RL, line 455: Just leave it here for potential consideration: this line alone spends ~2-4 ms of our performance budget (~16.7 ms in total if we're to maintain 60 fps). So we could potentially have a flag in device/index.js that gets set when one tool has enabled drawing / raycasting inside of it, and if no tool is enabled, we can skip this entire section.

ddgond: It's async, so unless I'm mistaken, this is just a 2-4ms latency between a mouse click and the tool receiving the event info. It should not be blocking the event loop/taking up our performance budget.

Steve-KX-RL: Oh make sense

PR #736, opened 2024-06-26
Re-enable frameNeedsToBeRendered in iOS app

hobinjk-ptc:
iOS sends a ton of matrix update messages which in turn redraw. The app can't keep up with the stream so this reinstates previous behavior of discarding updates if a new frame hasn't been rendered yet.


benptc (approved):
Seems like a good idea. I know this was causing problems for Rik but we can always dig into the root cause if/when we go further into WebXR

PR #740, opened 2024-07-01
Added support to read new GS Segmentation ply data

Steve-KX-RL:
1. Updated training pipeline to incorporate [SAGA](https://github.com/Jumpat/SegAnyGAussians) 3D segmentation data into .ply file.
2. Updated .ply / splat file parser to be compatible with the new .ply file structure
3. Added "segmentation mode" besides regular color mode to the splat viewer, triggered by keyboard "J"
4. Visualized segmentation cluster center points in the splat viewer. Point size corresponds to cluster size (bigger point means there are more splats in this cluster / category), and hovering over the points gives cluster label.
Example screenshot & video:
GS 3D Segmentation GUI

https://github.com/ptcrealitylab/vuforia-spatial-toolbox-userinterface/assets/111802681/df8a24af-a107-4dc8-80ab-8ff8533182a4

Will work with Dan to update the training pipeline automation. For now, if you want to try it out, see Teams chat for an example new version of the reality lab splat file.

### DIFF
--- a/src/splatting/Splatting.js
+++ b/src/splatting/Splatting.js
@@ -14,6 +14,8 @@ let gsContainer;
 let gsSettingsPanel;
 let isGSRaycasting = false;
 
+let labelInitialized = false;
+
 /** iPhoneVerticalFOV, projectionMatrixFrom(), makePerspective() come from desktopAdapter in remote operator addon. */
 
 const iPhoneVerticalFOV = 41.22673; // https://discussions.apple.com/thread/250970597
@@ -137,6 +139,15 @@ function multiply4(a, b) {
     ];
 }
 
+function multiply4v(m, v) {
+    return [
+        v[0] * m[0] + v[1] * m[4] + v[2] * m[8] + v[3] * m[12],
+        v[0] * m[1] + v[1] * m[5] + v[2] * m[9] + v[3] * m[13],
+        v[0] * m[2] + v[1] * m[6] + v[2] * m[10] + v[3] * m[14],
+        v[0] * m[3] + v[1] * m[7] + v[2] * m[11] + v[3] * m[15]
+    ]
+}
+
 function invert4(a) {
     let b00 = a[0] * a[5] - a[1] * a[4];
     let b01 = a[0] * a[6] - a[2] * a[4];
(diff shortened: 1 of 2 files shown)


src/splatting/Splatting.js
```diff
@@ -196,13 +207,15 @@ function ApplyTransMatrix(sourceMatrix, transMatrix, scaleF)
 function createWorker(self) {
     let buffer;
     let vertexCount = 0;
+    let doneLoading = false;
     let viewProj;
     // 6*4 + 4 + 4 = 8*4
     // XYZ - Position (Float32)
     // XYZ - Scale (Float32)
     // RGBA - colors (uint8)
     // IJKL - quaternion/rot (uint8)
```
hobinjk-ptc, line 216: should update this comment with documentation about the cluster label (4 uint8s?)

src/splatting/Splatting.js
```diff
@@ -254,33 +267,65 @@ function createWorker(self) {
...
-            texdata_c[4 * (8 * i + 7) + 1] = u_buffer[32 * i + 24 + 1];
-            texdata_c[4 * (8 * i + 7) + 2] = u_buffer[32 * i + 24 + 2];
-            texdata_c[4 * (8 * i + 7) + 3] = u_buffer[32 * i + 24 + 3];
+            texdata_c[4 * (8 * i + 7) + 0] = u_buffer[36 * i + 24 + 0];
+            texdata_c[4 * (8 * i + 7) + 1] = u_buffer[36 * i + 24 + 1];
+            texdata_c[4 * (8 * i + 7) + 2] = u_buffer[36 * i + 24 + 2];
+            texdata_c[4 * (8 * i + 7) + 3] = u_buffer[36 * i + 24 + 3];
+
+            // cx, cy, cz, label (cluster center (x, y, z) and splat label/category)
+            // todo Steve: cx, cy, cz is read as Uint8 , therefore not accurate, need to find a way to read / parse & store them as actual Float32 values, just like x, y, z.
```
hobinjk-ptc, line 290: Where are they written as uint8 instead of f32? I think the comment would be clearer if it pointed to the culprit

src/splatting/Splatting.js
```diff
@@ -445,7 +497,9 @@ function createWorker(self) {
         // XYZ - Scale (Float32)
         // RGBA - colors (uint8)
         // IJKL - quaternion/rot (uint8)
```
hobinjk-ptc, line 499: As above, should keep this line of comments up-to-date with the expected format

hobinjk-ptc (approved):
Overall LGTM! Excited to try this out

hobinjk-ptc:
A nice stretch goal to have would be to support the old format (or make a tool that adds missing cluster info to the old data instead of supporting both in the parser) so we don't have to re-train all the old splats

src/splatting/Splatting.js
Steve-KX-RL: Running out of space to store them😆 I only have 4 x uint8's left and I thought recovering a f32 from uint8 was possible somehow. But later I realized we actually don't have to store cx/cy/cz in each splats at all. I'll get rid of them and save some space for smth else we might store later.

PR #745, opened 2024-07-12
Capture canvas screenshot (for camera perspective memory bar)

benptc:
Goes with: https://github.com/ptcrealitylab/vuforia-spatial-remote-operator-addon/pull/202, to enable getting a thumbnail image of the scene. Just doing canvas.toDataURL at arbitrary moments elsewhere in the code may result in an empty image in between canvas draw calls.

camera-position-memory-bar-v5

### DIFF
--- a/src/gui/threejsScene.js
+++ b/src/gui/threejsScene.js
@@ -14,6 +14,7 @@ import AnchoredGroup from "./scene/AnchoredGroup.js";
 import { WebXRCamera, DefaultCamera, LayerConfig } from "./scene/Camera.js";
 import { Renderer } from "./scene/Renderer.js";
 import {setMatrixFromArray} from "./scene/utils.js";
+import { getPendingCapture } from './sceneCapture.js';
 
 (function(exports) {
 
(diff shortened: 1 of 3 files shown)


src/gui/threejsScene.js
```diff
@@ -1244,6 +1247,52 @@ import {setMatrixFromArray} from "./scene/utils.js";
         return new THREE.Vector3(x, y, z);
     }
 
+    let pendingCapture = null;
+
+    function capture(outputWidth, outputHeight, compressionOptions) {
```
hobinjk-ptc, line 1252: Should document this function's role and parameters

src/splatting/Splatting.js
```diff
@@ -1176,6 +1177,52 @@ async function main(initialFilePath) {
     }
 }
 
+let pendingCapture = null;
+
+function capture(outputWidth, outputHeight, compressionOptions) {
```
hobinjk-ptc, line 1182: See previous comment, also not ideal to have this level of repeated code but I guess it's okay for a little feature like this

hobinjk-ptc (approved):
Excited for the return of the memory bar! 😆

src/splatting/Splatting.js
Steve-KX-RL: 👆agreed. Maybe if we have time to revisit this later? Since it's a small feature.

Steve-KX-RL (approved):
LGTM! Thx for this feature!

dangond-ptc (approved):
LGTM!

src/splatting/Splatting.js
benptc: I'll see if there's a sensible way for me to refactor this before I merge the PR... 🤔

PR #749, opened 2024-07-18
Table view

dangond-ptc:
Adds a table view to the motion studies, launchable from the analytics settings menu. Includes ability to select different steps from the table view as well as different joints to draw spaghetti diagrams for. Also includes changes to the pop up menus that allow for maximizing them (The Analytics Settings menu does not yet benefit from this as it doesn't currently use the new DraggableMenu class, though it doesn't benefit from the feature either since the menu fits in a small box).


hobinjk-ptc (approved):
Molto bene!

PR #760, opened 2024-08-06
Fixed GS raycasting cursor jumping around bug + all math functions refactor

Steve-KX-RL:
Old GS raycasting method has a fundamental flaw which makes calculating cursor world position dependent on screen aspect ratio, specifically only accurate when screen is 1:1. Now introducing a new way to get raycast position. We use the camera space depth to compute the offset between camera and raycast position, and converting it back to world space. The result is a smooth & accurate spatial cursor, always at the center of mouse cursor, regardless of the screen aspect ratio.

https://github.com/user-attachments/assets/4642e6dc-e7f5-4b1c-8ce6-7d4f656698a6

### DIFF
--- a/src/splatting/Splatting.js
+++ b/src/splatting/Splatting.js
@@ -8,6 +8,7 @@
 import * as THREE from '../../thirdPartyCode/three/three.module.js';
 import GUI from '../../thirdPartyCode/lil-gui.esm.js';
 import { getPendingCapture } from '../gui/sceneCapture.js';
+import { remap } from "../utilities/MathUtils.js";
 
 let gsInitialized = false;
 let gsActive = false;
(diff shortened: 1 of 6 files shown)


src/splatting/Splatting.js
```diff
@@ -885,6 +889,34 @@ async function main(initialFilePath) {
...
+        return -Math.pow(r - 1, a) + 1;
+    }
+
+    const _remap = (x, lowIn, highIn, lowOut, highOut) => {
+        return lowOut + (highOut - lowOut) * remap01(x, lowIn, highIn);
+    }
+
+    const _remapCurve = (x, lowIn, highIn, lowOut, highOut) => {
+        return lowOut + (highOut - lowOut) * remap01Curve2(x, lowIn, highIn);
+    }
```
benptc, line 918: Would it be possible to refactor these common math functions into some kind of utils file, and import them where needed? I notice that similar/same functions to these are also declared in spatialCursor, spatialArrow, and spatialIndicator, which makes me think they could be refactored to a common location

benptc (approved):
Works well for me! Just left one comment about code duplication

src/splatting/Splatting.js
Steve-KX-RL: Good idea. Will create a math.js and refactor all repeated functions into it.

PR #763, opened 2024-08-08
improve the window.resize event to trigger if userinterface is in an iframe whose size gets changed suddenly

benptc:
This fixes issues where the drawing tool and thingview tool renderers' sizes don't update if you suddenly open/close the developer panel or the Teams chat window. I want to merge this in ASAP so that the upcoming live demo doesn't require users to resize their windows after adding tools.

### DIFF
--- a/src/device/layout.js
+++ b/src/device/layout.js
@@ -93,7 +93,6 @@ createNameSpace('realityEditor.device.layout');
         // (for example, if you open Developer Tools panel, or if while on the stage in Teams you open/close the chat)
         const resizeObserver = new ResizeObserver(entries => {
             for (let entry of entries) {
-                console.log('ResizeObserver resized (draw):', entry.contentRect.width, entry.contentRect.height);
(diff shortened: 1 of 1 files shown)


src/device/layout.js
```diff
@@ -88,12 +88,22 @@ createNameSpace('realityEditor.device.layout');
             toolSubscriptions[fullMessageContent.frame] = true;
         });
 
+        // Instead of using `window.addEventListener('resize', ...)`, we use a ResizeObserver, because the window
+        // `resize` event doesn't trigger under some circumstances when the document suddenly changes size
+        // (for example, if you open Developer Tools panel, or if while on the stage in Teams you open/close the chat)
+        const resizeObserver = new ResizeObserver(entries => {
```
hobinjk-ptc, line 94: Hooray for modern APIs that work better!

hobinjk-ptc (approved):

PR #764, opened 2024-08-08
Window resize workaround

benptc:
This feels too dubious to me, and I haven't actually tested if it works in Teams, but... hmm, maybe it's worth a try

### DIFF
--- a/src/device/layout.js
+++ b/src/device/layout.js
@@ -98,16 +98,25 @@ createNameSpace('realityEditor.device.layout');
         });
         resizeObserver.observe(document.body);
 
+        setInterval(() => {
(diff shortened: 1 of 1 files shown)


benptc:
Unfortunately this didn't even seem to fix it 100%, though it works better than it used to in Teams. Usually works, but that 10% of the time that it doesn't still bothers me. I recorded a video with some captions of what's happening: https://ptccloud-my.sharepoint.com/:v:/g/personal/breynolds_ptc_com/EWqhNomUL2FOt61p_MB_NcYBK9wnUMb61Nbmv3hwHQeong?e=IhKL3h, for future reference.

PR #767, opened 2024-08-20
Error handling around navmesh write to local storage

benptc:
Prevents the session from breaking when trying to load an unusually large area target, e.g. this scan of the 17th floor

Screenshot 2024-08-20 at 11 47 18 AM


dangond-ptc (approved):
Good catch and good change!

PR #768, opened 2024-08-20
Table View Coloring and MURI images

dangond-ptc:
Added coloring to the analytics table view and explanatory images to the MURI columns in the table view
Screenshot 2024-08-20 at 1 53 42 PM


src/utilities/TableView.js
```diff
@@ -90,10 +106,13 @@ export class TableView {
...
             }
-            const cell = e.target;
+            let cell = e.target;
+            if (cell.tagName === 'IMG') { // Handle clicks on header images
+                cell = cell.parentNode;
+            }
             const cellIndex = cell.cellIndex;
             const rowIndex = cell.parentNode.rowIndex;
-
+
```
hobinjk-ptc, line 115: nit: trailing whitespace

hobinjk-ptc (approved):

benptc (approved):
Code LGTM, looks good to merge.

One small thing from a UX perspective is that for someone unfamiliar with MURI the value of the numbers in the boxes don't have a clear explanation as they are unitless and have no min/max communicated in the UI. The red/yellow/green and the pictures is super nice and clears up the big picture of what they mean, but I wonder in future if it would be good to also have a little tooltip explanation or a `(?)` link you can click on with a reference to how the MURI scoring system works.

PR #770, opened 2024-08-23
Optimize network bandwidth used while in avatar following mode

benptc:
Reduces the rate that you send your avatar position to the server; uses 10 fps normally, but reduces it to 1 fps if you're following another user's avatar (referred to in the code as `lockOnMode`). This is to prevent everyone from spamming the server at the same time while a large group is following a single avatar that's moving around.

Works best in combination with https://github.com/ptcrealitylab/vuforia-spatial-edge-server/pull/1125


dangond-ptc (approved):
LGTM!

PR #779, opened 2024-09-11
hide UI that says X users are following you, until your avatar has finished initializing

benptc:
Fixes bug where `myId` starts out null, so when the page is loading it might look like other users are following you if they aren't following anyone (because their lock on ID is also null).

Screenshot 2024-09-11 at 12 35 32 PM


Steve-KX-RL (approved):
Nice catch!

PR #803, opened 2024-10-24
Invert video player shader

hobinjk-ptc:
For complex iOS-side reasons the videos are upside-down.

### DIFF
--- a/src/spatialCapture/Shaders.js
+++ b/src/spatialCapture/Shaders.js
@@ -32,6 +32,7 @@ uniform vec2 principalPoint;
 const float pointSizeBase = 0.0;
 
 uniform bool useFarDepth;
+uniform bool inverted;
 
 varying vec2 vUv;
 varying vec4 pos;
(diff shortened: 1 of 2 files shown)


src/spatialCapture/Shaders.js
```diff
@@ -314,7 +320,8 @@ export function createPointCloudMaterial(texture, textureDepth, shaderMode, bord
             viewAngleSimilarity: { value: 0.0 },
             viewPositionSimilarity: { value: 0.0 },
 			// used to erase depth in the pre render
-            useFarDepth: {value: false}
+            useFarDepth: {value: false},
```
dangond-ptc, line 323: Is useFarDepth used?

src/spatialCapture/Shaders.js
```diff
@@ -314,7 +320,8 @@ export function createPointCloudMaterial(texture, textureDepth, shaderMode, bord
...
+            inverted: {value: false}
```
dangond-ptc, line 324: Do we not want it always set to true?

src/spatialCapture/Shaders.js
hobinjk-ptc: It's used in vuforia-spatial-toolbox-userinterface/src/spatialCapture/CameraVisPatch.js

src/spatialCapture/Shaders.js
hobinjk-ptc: The live video is not inverted, keeps it interesting :D

dangond-ptc (approved):

PR #807, opened 2024-10-29
Respect RegionCard startTime and endTime where possible

hobinjk-ptc:
Ends previous behavior of cards overwriting startTime and endTime from pose data. Also fixes a bug where changes to regioncard start/end wouldn't cause themselves to persist.


dangond-ptc (approved):
LGTM, when should we use scheduleWriteMotionStudyData over the standard writeMotionStudyData?

hobinjk-ptc:
When you don't care about it happening instantly

PR #811, opened 2024-10-31
Extend sidebar to full height; Scooch timeline over to not overlap

hobinjk-ptc:


benptc (approved):
Tested it out, new layout LGTM!

